PR #16, opened 2020-12-09
14.0 mig hr holidays public

JunglistMNSQ:


pedrobaeza:
There's a temporary Travis glitch that we are trying to fix in https://github.com/OCA/maintainer-tools/pull/483, but you will need to rebase when fixed.

JunglistMNSQ:
> There's a temporary Travis glitch that we are trying to fix in [OCA/maintainer-tools#483](https://github.com/OCA/maintainer-tools/pull/483), but you will need to rebase when fixed.

Is It normally that this pull request has been checking for a half an hour already?

PR #17, opened 2020-12-23
[13.0][ADD] hr_holidays_natural_period: Add natural day in request unit leave types

victoralmau:
Add natural day in request unit leave types

@Tecnativa TT27491


hr_holidays_natural_period/models/hr_leave.py
```diff
@@ -0,0 +1,22 @@
...
+class HrLeave(models.Model):
+    _inherit = "hr.leave"
+
+    def _get_number_of_days(self, date_from, date_to, employee_id):
+        if (
+            self.holiday_status_id
+            and self.holiday_status_id.request_unit == "natural_day"
+            and date_from
+            and date_to
+        ):
```
pedrobaeza: Why do you need such complex condition?
```suggestion
        if self.holiday_status_id.request_unit == "natural_day":
```

hr_holidays_natural_period/models/resource_calendar.py
```diff
@@ -0,0 +1,48 @@
...
+                                    datetime.combine(dt_item, time.max).replace(
+                                        tzinfo=tz
+                                    ),
+                                    self.env["resource.calendar.leaves"],
+                                )
+                            )
+                            dt_item = dt_item + relativedelta(days=1)
+                        intervals[resource.id] = Intervals(new_intervals)
+                else:
+                    intervals[resource] = Intervals([])
```
pedrobaeza: Do you need this `else` part? It should be already filled by super.

hr_holidays_natural_period/models/resource_calendar.py
```diff
@@ -0,0 +1,48 @@
...
+            start_dt=start_dt, end_dt=end_dt, resources=resources, domain=domain, tz=tz
+        )
+        if self.env.context.get("natural_period"):
+            for resource in resources:
+                if resource:
+                    interval_resource = intervals[resource.id]
+                    if len(interval_resource._items) > 0:
+                        dt_item = start_dt.date()
+                        new_intervals = []
+                        while dt_item != end_dt.date():
```
pedrobaeza: A safer option is to compare this way:
```suggestion
                        while dt_item < end_dt.date():
```

hr_holidays_natural_period/models/resource_calendar.py
```diff
@@ -0,0 +1,48 @@
...
+                            new_intervals.append(
+                                (
+                                    datetime.combine(dt_item, time.min).replace(
```
pedrobaeza: I'm not sure about using this, as it can be conflicting due to timezone, showing holidays from previous/next days to the rest if in different zones. Real case: A leave from someone in Portugal for the day 2 of February, is shown in the calendar as being from 1 to 2 of February for Spanish HR managers.

Possibilities:

1. Put 12:00 as compromise solution (for the most extreme tz difference).
2. Take the first real calendar attendance at copy the starting hour.

The same can be applied to the ending hour.

hr_holidays_natural_period/models/resource_calendar.py
```diff
@@ -0,0 +1,48 @@
...
+                                        tzinfo=tz
+                                    ),
+                                    datetime.combine(dt_item, time.max).replace(
+                                        tzinfo=tz
+                                    ),
+                                    self.env["resource.calendar.leaves"],
```
pedrobaeza: Shouldn't this be an instance of `resource.calendar.attendances`?

pedrobaeza (changes requested):

pedrobaeza:
Please check if there's a method in resource that gets you the suitable schedule. Anyway, right now is not correct, as you are taking the first resource.calendar.attendance as start and end, but if you have for example, monday from 8:00 to 12:00 and monday from 14:00 to 18:00, it will take 12:00, and not 18:00 as ending hour.

victoralmau:
Timezone is defined in leave (https://github.com/odoo/odoo/blob/13.0/addons/hr_holidays/models/hr_leave.py#L141)
`hr_holidays` addon use it https://github.com/odoo/odoo/blob/13.0/addons/hr_holidays/models/hr_leave.py#L472 and in function https://github.com/odoo/odoo/blob/13.0/addons/resource/models/resource.py#L344 exist timezone.

Now i have changed function to get first attendance correctly and add test about it.

pedrobaeza (changes requested):
I'm afraid this still doesn't serve. Several comments:

- Resource calendar has also tz, so you should handle both tz parts (but be sure that the `_attendance_intervals_batch` should return localized intervals, and not naive (without tz) ones.
- Taking the maximum date doesn't serve, as it can vary from each day, and having monday with 8 hours from 8:00 to 16:00, and friday from 10:00 to 18:00, so you will put in the leave from 08:00 to 18:00, which is not correct. You have to take the first calendar.attendance, and then find the possible `afternoon` attendance for that day, taking the ending hour. I insist that there should be a method for getting this interval already for not reinventing the wheel.

pedrobaeza (approved):

hr_holidays_natural_period/readme/USAGE.rst
```diff
@@ -0,0 +1,9 @@
+For using natural period on leaves:
+
+#. Go to *Leaves > Dashboard*.
+#. Select dragging on the calendar the days you want to be on leave, or go
+   to the form view for selecting start and end dates.
+#. Select the proper "Leave Type" that has "Natural day" selected in "Request unit".
```
chienandalu, line 6: Testing functionally I found a little UI problem: If you change dates, the amount of days change, but if you change days, the dates don't adjust, so you could get whatever amount of days for whatever range of dates

victoralmau: It's interesting UI case but, i had checked similar in runbot 13.0 with other type and it work equal (change duration days NOT re-calculate dates).

chienandalu: Mmm, intersting and confusing. Does it make any sense?

victoralmau: IMO it's a bug because it's need to auto-adjust date end (according to number of days) but maybe it's a question to odoo, right? @pedrobaeza , what do you think about it?

chienandalu: Yeah, I'm thinking in an Odoo issue :slightly_smiling_face: 

The easy fix to prevent this would be to make the field readonly in view anyway

pedrobaeza: That's something that shouldn't be done in this module though, so let's not block it here.

chienandalu (approved):

PR #20, opened 2021-01-20
[13.0][FIX] hr_holidays_public: Union some models (hr.holidays.public.line + resource.calendar.leaves)

victoralmau:
In some cases error appear error to union some models (`hr.holidays.public.line` + `resource.calendar.leaves`)
It's only happen when exists holidays + leaves (`hr_attendance_report_theoretical_time` addon)
Solution is define index_2 (anyone) in all intervals created because index_2 is not used

@Tecnativa TT27757

### DIFF
--- a/hr_holidays_public/models/resource_calendar.py
+++ b/hr_holidays_public/models/resource_calendar.py
@@ -42,12 +42,20 @@ class ResourceCalendar(models.Model):
             lines = HrHolidaysPublic.get_holidays_list(
                 day.year, employee_id=employee_id
             )
+            # In some cases, an error appears about mixing 2 models
+            # (hr.holidays.public.line + resource.calendar.leaves)
+            # in _leave_intervals or _leave_intervals_batch functions
+            # It only happen when both holidays and leaves exist.
+            # The solution is to pass an empty leave in the tuple instead
+            # of the public holiday line record, as this element has no
+            # further use except the union operation.
+            resource_leave_model = self.env["resource.calendar.leaves"]
             for line in lines:
                 leaves.append(
                     (
                         datetime.combine(line.date, time.min).replace(tzinfo=tz),
                         datetime.combine(line.date, time.max).replace(tzinfo=tz),
-                        line,
+                        resource_leave_model,
                     )
                 )
         return Intervals(leaves)


hr_holidays_public/models/resource_calendar.py
```diff
@@ -42,12 +42,18 @@ def _public_holidays_leave_intervals(self, start_dt, end_dt, employee_id, tz):
             lines = HrHolidaysPublic.get_holidays_list(
                 day.year, employee_id=employee_id
             )
+            # In some cases error appear error to union some models
+            # (hr.holidays.public.line + resource.calendar.leaves)
+            # in _leave_intervals_batch or _leave_intervals_batch functions
+            # It's only happen when exists holidays + leaves
+            # Solution is define index_2 in all intervals created
+            # because index_2 is not used
```
pedrobaeza:
```suggestion
            # In some cases, an error appears about mixing 2 models
            # (hr.holidays.public.line + resource.calendar.leaves)
            # in _leave_intervals_batch or _leave_intervals_batch functions
            # It only happen when both holidays and leaves exist.
            # The solution is to pass an empty leave in the tuple instead
            # of the public holiday line record, as this element has no
            # further use except the union operation.
```

pedrobaeza (approved):

hr_holidays_public/models/resource_calendar.py
```diff
@@ -42,12 +42,19 @@ def _public_holidays_leave_intervals(self, start_dt, end_dt, employee_id, tz):
...
+            # The solution is to pass an empty leave in the tuple instead
+            # of the public holiday line record, as this element has no
+            # further use except the union operation.
             for line in lines:
                 leaves.append(
                     (
                         datetime.combine(line.date, time.min).replace(tzinfo=tz),
                         datetime.combine(line.date, time.max).replace(tzinfo=tz),
-                        line,
+                        self.env["resource.calendar.leaves"],
```
Tardo: Can put this out of the loop to access the item only once.

Tardo (approved):
LGTM, code review. Only a non-blocking comment.

hr_holidays_public/models/resource_calendar.py
```diff
@@ -42,12 +42,20 @@ def _public_holidays_leave_intervals(self, start_dt, end_dt, employee_id, tz):
             lines = HrHolidaysPublic.get_holidays_list(
                 day.year, employee_id=employee_id
             )
+            # In some cases, an error appears about mixing 2 models
+            # (hr.holidays.public.line + resource.calendar.leaves)
+            # in _leave_intervals_batch or _leave_intervals_batch functions
```
pedrobaeza: I'm seeing now that it's the same function twice. Do you mean another?

victoralmau: It's correct, 2 different functions are needed.
In 14.0 `_leave_intervals` remove, https://github.com/OCA/hr-holidays/blob/13.0/hr_holidays_public/models/resource_calendar.py#L56

pedrobaeza: Then you should put:
```suggestion
            # in _leave_intervals or _leave_intervals_batch functions
```

PR #22, opened 2021-04-14
[13.0][FIX] hr_holidays_public: Fix number_of_days in leaves when contain holiday in leave and hour start from resource calendar in 00:01

victoralmau:
Fix `number_of_days` in leaves when contain holiday in leave and hour start from resource calendar in 00:01.
It's a problem according to timezone.

Steps to reproduce:
- Create 2021-05-01 in public holidays.
- Create new working hours according to Abigail Peterson called "_Tuesday - Saturday 00:01 - 08:01_" and set details.
- Go to Time off > Managers > To aprrove > Time Off and create new records:
- Record A: Michel Admin, Paid Time Off, from 2021-04-27 to 2021-05-01.
- Record B: Abigail Peterson, Paid Time Off, from 2021-04-27 to 2021-05-01.

In some examples set duration to 5 days.

@Tecnativa TT29189

### DIFF
--- a/hr_holidays_public/models/hr_holidays_public.py
+++ b/hr_holidays_public/models/hr_holidays_public.py
@@ -1,6 +1,7 @@
 # Copyright 2015 2011,2013 Michael Telahun Makonnen <mmakonnen@gmail.com>
 # License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
 
+import datetime
 from datetime import date
 
 from odoo import SUPERUSER_ID, _, api, fields, models
@@ -55,15 +56,23 @@ class HrHolidaysPublic(models.Model):
 
     @api.model
     @api.returns("hr.holidays.public.line")
-    def get_holidays_list(self, year, employee_id=None):
+    def get_holidays_list(
+        self, year=None, start_dt=None, end_dt=None, employee_id=None
+    ):
         """
         Returns recordset of hr.holidays.public.line
         for the specified year and employee
-        :param year: year as string
+        :param year: year as string (optional if start_dt and end_dt defined)
+        :param start_dt: start_dt as date
+        :param end_dt: end_dt as date
         :param employee_id: ID of the employee
         :return: recordset of hr.holidays.public.line
         """
-        holidays_filter = [("year", "=", year)]
+        if not start_dt and not end_dt:
+            start_dt = datetime.date(year, 1, 1)
+            end_dt = datetime.date(year, 12, 31)
+        years = list(range(start_dt.year, end_dt.year + 1))
+        holidays_filter = [("year", "in", years)]
         employee = False
         if employee_id:
             employee = self.env["hr.employee"].browse(employee_id)
@@ -88,6 +97,8 @@ class HrHolidaysPublic(models.Model):
             ]
         else:
             states_filter.append(("state_ids", "=", False))
+        states_filter.append(("date", ">=", start_dt))
+        states_filter.append(("date", "<=", end_dt))
         hhplo = self.env["hr.holidays.public.line"]
         holidays_lines = hhplo.search(states_filter)
         return holidays_lines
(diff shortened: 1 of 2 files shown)


hr_holidays_public/models/hr_holidays_public.py
```diff
@@ -55,42 +56,37 @@ def name_get(self):
 
     @api.model
     @api.returns("hr.holidays.public.line")
-    def get_holidays_list(self, year, employee_id=None):
+    def get_holidays_list(self, date_from, date_to, employee_id=None):
```
pedrobaeza: Please don't change API. You can make `year` a keyword argument, and add `date_from` and `date_to` as other keyword arguments, so they are optional, and use the most convenient one.

hr_holidays_public/models/resource_calendar.py
```diff
@@ -15,77 +11,33 @@
...
-        return res
-
-    def _leave_intervals_batch(
+    def _attendance_intervals_batch_exclude_public_holidays(
+        self, start_dt, end_dt, intervals, resources, tz, lines_date
+    ):
+        for resource in resources:
+            interval_resource = intervals[resource.id]
+            attendances = []
+            if len(interval_resource._items) > 0:
```
pedrobaeza: This line is not needed.

pedrobaeza (changes requested):

hr_holidays_public/models/resource_calendar.py
```diff
@@ -15,77 +11,30 @@
...
-            )
-            res = res | public_holidays
-        return res
-
-    def _leave_intervals_batch(
+        for resource in resources:
+            interval_resource = intervals[resource.id]
+            attendances = []
+            for attendance in interval_resource._items:
+                if not HrHolidaysPublic.is_public_holiday(
```
pedrobaeza: This seems less performant. I think it's still a good idea to extend `get_holidays_list` (or create another one, but not duplicating code) for searching for a date range.

hr_holidays_public/models/resource_calendar.py
```diff
@@ -15,77 +11,42 @@
...
-                        datetime.combine(line.date, time.max).replace(tzinfo=tz),
-                        resource_leave_model,
+        list_by_years = {}
+        for resource in resources:
+            interval_resource = intervals[resource.id]
+            attendances = []
+            for attendance in interval_resource._items:
+                year_item = attendance[0].date().year
+                if year_item not in list_by_years:
+                    list_by_years[year_item] = HrHolidaysPublic.get_holidays_list(
```
pedrobaeza: You are still calling too much this method (one per resource attendance). As previously said, it's not a bad idea to have a method to retrieve by date range the public holidays. My only comment was not to change the API of the current method, creating other method in a way that reuse the same code (for example leaving the other method signature and calling a new one with dates having from 01/01 to 31/12), or putting all arguments (year, date_start and date_end) in the same method as keyword ones.

hr_holidays_public/models/hr_holidays_public.py
```diff
@@ -53,6 +53,14 @@ def name_get(self):
             result.append((rec.id, rec.display_name))
         return result
 
+    @api.model
+    @api.returns("hr.holidays.public.line")
+    def get_holidays_list_by_dates(self, start_dt, end_dt, employee_id=None):
+        holidays_lines = self.env["hr.holidays.public.line"]
```
pedrobaeza: My idea was the contrary: the base method to be the date range method, and this is not correct, as your adding more lines than expected.

hr_holidays_public/models/hr_holidays_public.py
```diff
@@ -55,15 +55,17 @@ def name_get(self):
 
     @api.model
     @api.returns("hr.holidays.public.line")
-    def get_holidays_list(self, year, employee_id=None):
+    def get_holidays_list_by_year(self, years, employee_id=None):
```
pedrobaeza: You are still changing the API... Please check all my comments and my explanations, and ask me for any doubt, but let's try to avoid loops.

pedrobaeza (approved):

pedrobaeza:
Thanks for the effort!

/ocabot merge patch

pedrobaeza:
Now we need to fw-port it to v14

PR #30, opened 2021-09-29
[14.0] hr_holidays_public: add public holidays as unusual days

petrus-v:
in order to make a grey/disabled date in the calendar view

### DIFF
--- a/hr_holidays_public/models/hr_leave.py
+++ b/hr_holidays_public/models/hr_leave.py
@@ -42,6 +42,18 @@ class HrLeave(models.Model):
                     ("year_id.country_id", "=", country_id),
                 ]
             )
+        state_id = self.env.user.employee_id.address_id.state_id.id
+        if not state_id:
+            country_id = self.env.company.state_id.id or False
+        if state_id:
+            domain.extend(
+                [
+                    "|",
+                    ("state_ids", "in", [state_id]),
+                    ("state_ids", "=", False),
+                ]
+            )
+
         public_holidays = self.env["hr.holidays.public.line"].search(domain)
         for public_holiday in public_holidays:
             res[fields.Date.to_string(public_holiday.date)] = True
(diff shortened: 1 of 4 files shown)


pedrobaeza:
Public holidays may depend on the user and their location (country and state), so we should restrict it. It's the call to this method linked to one user? Now that currently a calendar event is created for the public holiday.

petrus-v:
> Public holidays may depend on the user and their location (country and state), so we should restrict it. It's the call to this method linked to one user? Now that currently a calendar event is created for the public holiday.

Good catch !

[hr holliday module consider current user](https://github.com/OCA/OCB/blob/14.0/addons/hr_holidays/models/hr_leave.py#L1257) to display unusual days (like weekends according calendar resource model): for a manager that would validate holidays to an employee it will get it's own unusual days (not the one from the employee that he is checking).

So what i can do is similar: filter public holidays by country of the current user according country define on self.env.user.employee_id.address_id.country_id (with fallback to the current company country) if set assuming manager live in the same country as subordinates and add a warning with this limitation in readme.

pedrobaeza:
Great, we have a plan :wink:

hr_holidays_public/readme/DESCRIPTION.rst
```diff
@@ -1,4 +1,6 @@
-This module is a technical module to handle public holidays.
+This module handle public holidays.
```
pedrobaeza:
```suggestion
This module handles public holidays.
```

hr_holidays_public/models/hr_leave.py
```diff
@@ -18,3 +18,31 @@ def _get_number_of_days(self, date_from, date_to, employee_id):
...
+                (
+                    "date",
+                    "<",
+                    date_to,
+                )
+            )
+        country_id = self.env.user.employee_id.address_id.country_id.id
+        if not country_id:
+            country_id = self.env.company.country_id.id or False
+        if country_id:
```
pedrobaeza, line 37: Something similar should be done for the states.

pedrobaeza (approved):

AntoniRomera (approved):

AntoniRomera:
I see this as a good addition, good work.

PR #32, opened 2021-10-29
[15.0][REF][MIG]hr_holidays_validity_date

AaronHForgeFlow:
Refactor done so we cover the case where time off types do not require allocations. In that case we may want to restrict the dates by type, same as it was in previous versions.


AaronHForgeFlow:
Validity dates are now available in the allocation request so IMO this module is not needed anymore.

PR #36, opened 2021-12-16
[14.0][Fix] hr_holiday_public: Work time per day not excluding holidays

santostelmo:
### Context
The mehod `resource_calendar.list_work_time_per_day(self, from_datetime, to_datetime, calendar=None, domain=None)`  returns a list of tuples (day, hours). Working days and number of hours to work on that day.
The issue is that is not excluding public holidays.

Steps to reproduce:
- Please check the unit test => Without the fixing, `list_work_time_per_day` will return 3 working days when is suposed to be 2.

This has an impact for example when module `project_timesheet_holidays` is installed. Normally with that module timesheet lines of type holidays are created  when a `hr.leave` is validated. But with that bug, also timesheet lines are created in public holidays. The method `list_work_time_per_day` is not propagating context variables `exclude_public_holidays` and `employee_id`.

### DIFF
--- a/hr_holidays_public/models/hr_leave.py
+++ b/hr_holidays_public/models/hr_leave.py
@@ -8,6 +8,17 @@ from odoo import api, fields, models
 class HrLeave(models.Model):
     _inherit = "hr.leave"
 
+    def action_validate(self):
+        """Inject the needed context for excluding public holidays (if applicable) on the
+        actions derived from this validation. This is required for example for
+        `project_timesheet_holidays` for not generating the timesheet on the public holiday.
+        Unfortunately, no regression test can be added, being in a separate module."""
+        if self.holiday_status_id.exclude_public_holidays or not self.holiday_status_id:
+            self = self.with_context(
+                employee_id=self.employee_id.id, exclude_public_holidays=True
+            )
+        return super(HrLeave, self).action_validate()
+
     def _get_number_of_days(self, date_from, date_to, employee_id):
         if self.holiday_status_id.exclude_public_holidays or not self.holiday_status_id:
             instance = self.with_context(


ivantodorovich (approved):
LG

ivantodorovich:
> LG

Travis disagrees though

santostelmo:
The order of the dates comparision in the unit test is wrong.

hr_holidays_public/models/resource_mixin.py
```diff
@@ -0,0 +1,19 @@
+# Copyright 2021 Camptocamp - Telmo Santos
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
+
+
+from odoo import models
+
+
+class ResourceMixin(models.AbstractModel):
+    _inherit = "resource.mixin"
```
i-vyshnevska:
```suggestion
    _inherit = "hr.employee"
```
what about adding this change not globally to mixin just to employees for instance? 
mixin is used in other objects and this behavior may not be wanted there

santostelmo: Good idea. I agree

santostelmo:
@i-vyshnevska I updated the code according to you comment.

i-vyshnevska (approved):

hr_holidays_public/models/hr_employee.py
```diff
@@ -0,0 +1,19 @@
...
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
+
+
+from odoo import models
+
+
+class HrEmployee(models.Model):
+    _inherit = "hr.employee"
+
+    def list_work_time_per_day(
```
gurneyalex: the method is defined on resource.mixin, and this is the model you need to patch, otherwise all other models which inherit from the mixin will have the wrong method. 

See https://github.com/odoo/odoo/blob/14.0/addons/resource/models/resource_mixin.py#L167

gurneyalex (changes requested):

hr_holidays_public/models/hr_employee.py
santostelmo: In my first implementation I did the override of the method `list_work_time_per_day` in resource.mixin. But as suggested by @i-vyshnevska this change should be applied only for employees. mixin is used in other objects and this behavior may not be wanted there

JeremieJordan:
Functional test ok

hr_holidays_public/models/hr_employee.py
gurneyalex: then please add a comment in the source code explaining why you are doing this on that method and not on the mixin, and a unit test using another model inheriting from the mixin so that we get a clear information by reading the code that it is not an oversight.

i-vyshnevska: yep, it was my suggestion, https://github.com/OCA/hr-holidays/pull/36#discussion_r770707321 
work-center is not an employee why it should consider employee, as it OCA better to localize it just to employee

santostelmo: @gurneyalex I added the following comment:

> This method is originaly implemented in resource.mixin and for example,
> hr.employee and mrp.workcenter are inheriting that. We override it here, to pass
> context variables employee_id and exclude_public_holidays ONLY for instances
> of hr.employee (not for mrp.workcenter), so that public holidays are properly
> calculated.

santostelmo: @gurneyalex Regading the unit test where I should use a class inheriting from `resource.mixin` that is not `hr.employee`, I don't see how I can do such a test. For example `mrp.workcenter` is defined in mrp module and is not installed.

gurneyalex: @santostelmo  check the resource module tests: there is a ressource.test model you can use.

hr_holidays_public/tests/test_holidays_calculation.py
```diff
@@ -191,3 +191,37 @@ def test_number_of_hours_excluding_employee_2(self):
...
+                "employee_id": self.employee_2.id,
+            }
+        )
+        work_hours_data = self.employee_2.list_work_time_per_day(
+            leave_request.date_from, leave_request.date_to
+        )
+
+        self.assertEqual(len(work_hours_data), 2)
+
+    def test_list_work_time_per_day_on_resource_test_model(self):
```
santostelmo: @gurneyalex thanks for you advise to use `resource.test`. I created the unit test with that.

gurneyalex (approved):

hr_holidays_public/tests/test_holidays_calculation.py
```diff
@@ -191,3 +191,37 @@ def test_number_of_hours_excluding_employee_2(self):
 
         self.assertEqual(leave_request.number_of_days, 2)
         self.assertEqual(leave_request.number_of_hours_display, 16)
+
+    def test_list_work_time_per_day_excluding_employee_2(self):
+        self.holiday_type.request_unit = "hour"
+        leave_request = self.HrLeave.new(
```
pedrobaeza: Why are you creating a leave request if it's not used at the end?

pedrobaeza (changes requested):
Not sure if the problem is custom code that is directly calling the method `list_work_time_per_day`. I need steps to reproduce the problem in Odoo UI to validate this change.

hr_holidays_public/tests/test_holidays_calculation.py
santostelmo: The leave request is created in a period where there is public holidays. `list_work_time_per_day` should take in account the public holidays. 
The goal of this unit test is to show that `list_work_time_per_day returns` the correct output (2 working days).

pedrobaeza: As said, you are not using it in the unit test. If you remove it, and only pass directly the start and end dates, you will get the same result. Anyways, that's not the main concern. See my comment on the requested changes comment.

santostelmo:
> Not sure if the problem is custom code that is directly calling the method `list_work_time_per_day`. I need steps to reproduce the problem in Odoo UI to validate this change.

@pedrobaeza  steps to reproduce:

1. Install project_timesheet_holidays
2. Create a public holiday on 2021-12-25
3. Create a holiday from 2021-12-25 to 2021-12-25
4. Validate holiday

Result:
Timesheet line is created on 2021-12-25

Expected result:
No timesheet line is created on 2021-12-25 since it is a plublic holiday


When holiday is validated the following methods are called:
1. Validate timsheet - https://github.com/odoo/odoo/blob/4c42e43cdc35d46c37805b4ae9211f9ff1729174/addons/project_timesheet_holidays/models/hr_holidays.py#L60
2. Create timesheet lines https://github.com/odoo/odoo/blob/4c42e43cdc35d46c37805b4ae9211f9ff1729174/addons/project_timesheet_holidays/models/hr_holidays.py#L74 => Here `list_work_time_per_day` is called without context variable `exclude_public_holidays?

pedrobaeza:
Thanks for the explanation. Then I think you should override `_validate_leave_request` method and inject there the context, so any other possible module with similar problems is automatically fixed. Take also into account that the exclude public holidays is optional per leave type, so you mustn't assume always such thing.

santostelmo:
@pedrobaeza thanks for having spotted that. Indeed if I inject `exclude_public_holidays` in the method `list_work_time_per_day`, the possibilility to exclude public holidays will be lost.
So, as you suggested, I add the variable in the validation step.
https://github.com/odoo/odoo/blob/36d2643e19ef7868a475751b19305402028fb231/addons/hr_holidays/models/hr_leave.py#L990

To provide a unit test I don't see a way to do it since `project_timesheet_holidays` is not installed and what I would like to test is the correct creation to timesheet lines that is done here: https://github.com/odoo/odoo/blob/36d2643e19ef7868a475751b19305402028fb231/addons/project_timesheet_holidays/models/hr_holidays.py#L60

I implement that changes...

santostelmo:
@pedrobaeza changes done as mentioned in the previous comment.

hr_holidays_public/models/hr_leave.py
```diff
@@ -8,6 +8,13 @@
 class HrLeave(models.Model):
     _inherit = "hr.leave"
 
+    def action_validate(self):
+        if self.holiday_status_id.exclude_public_holidays or not self.holiday_status_id:
```
pedrobaeza, line 16: Put a docstring like this one for knowing the purpose of this overriding, or in the future, it may be removed if its goal is not understood:
```
"""Inject the needed context for excluding public holidays (if applicable) on the
actions derived from this validation. This is required for example for
`project_timesheet_holidays` for not generating the timesheet on the public holiday.
Unfortunately, no regression test can be added, being in a separate module.
"""
```

pedrobaeza:
And please include the steps to reproduce the problem and the rest of the explanation also in the message commit. Let's be generous on explanations, that are very useful later.

pedrobaeza:
And the last request would be to directly squash the commits, as the bot is not able to do it.

hr_holidays_public/models/hr_leave.py
santostelmo: Docstring added

pedrobaeza:
The expanded commit message is missing.

santostelmo:
> The expanded commit message is missing.

Not sure the about the expanded commit message but I changed it.

pedrobaeza:
The commit message should include the steps to reproduce the problem and the actions taken (and please don't include the version there. It should be only on the PR title):
```
[FIX] hr_holiday_public: Exclude public holidays validating leaves

Steps to reproduce the problem:
- Install project_timesheet_holidays
- Create a public holiday on 2021-12-25
- Create a time-off from 2021-12-25 to 2021-12-25
- Validate time-off

Result:
Timesheet line is created on 2021-12-25

Expected result:
No timesheet line is created on 2021-12-25 since it is a public holiday.

Injecting the proper context when validating the leave avoids the problem.
```

santostelmo:
Thanks for your help @pedrobaeza. I wrote the extended commit

pedrobaeza (approved):

PR #37, opened 2022-01-05
[13.0][FIX] hr_holidays_natural_period: Fix _exist_interval_in_date() function + Impove tests.

victoralmau:
Previously, the `_exist_interval_in_date()` function always returned `False`, which always created a new interval.
Although this was totally incorrect, it was detected in a real use case when there were working days and holidays in the date range causing incorrect overtime.

@Tecnativa TT33779

### DIFF
--- a/hr_holidays_natural_period/models/resource_calendar.py
+++ b/hr_holidays_natural_period/models/resource_calendar.py
@@ -16,7 +16,7 @@ class ResourceCalendar(models.Model):
 
     def _exist_interval_in_date(self, intervals, date):
         for interval in intervals:
-            if interval[0].date == date:
+            if interval[0].date() == date:
                 return True
         return False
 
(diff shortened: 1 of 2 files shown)


victoralmau:
There is a strange error with pre-commit that I don't understand, any ideas @joao-p-marques ?

joao-p-marques:
> There is a strange error with pre-commit that I don't understand, any ideas @joao-p-marques ?

The error is quite unclear to me as well… As the Travis CI is being deprecated, I think this is a good opportunity to migrate to the new Github Actions. I will open the PR to the main branch and, once merged, you can rebase here.

pedrobaeza (approved):

pedrobaeza:
Fast-tracking as obvious patch:

/ocabot merge patch

PR #46, opened 2022-06-14
[15.0][IMP] hr_holidays_public: use api.model_create_multi

gurneyalex:
decorate hr.holidays.public.line::create with api.model_create_multi

### DIFF
--- a/hr_holidays_public/models/hr_holidays_public.py
+++ b/hr_holidays_public/models/hr_holidays_public.py
@@ -213,12 +213,13 @@ class HrHolidaysPublicLine(models.Model):
             if rec.meeting_id:
                 rec.meeting_id.write(rec._prepare_holidays_meeting_values())
 
-    @api.model
-    def create(self, values):
-        res = super().create(values)
-        res.meeting_id = self.env["calendar.event"].create(
-            res._prepare_holidays_meeting_values()
-        )
+    @api.model_create_multi
+    def create(self, vals_list):
+        res = super().create(vals_list)
+        for record in res:
+            record.meeting_id = self.env["calendar.event"].create(
+                record._prepare_holidays_meeting_values()
+            )
         return res
 
     def unlink(self):


hr_holidays_public/models/hr_holidays_public.py
```diff
@@ -213,11 +213,13 @@ def _update_calendar_event(self):
             if rec.meeting_id:
                 rec.meeting_id.write(rec._prepare_holidays_meeting_values())
 
-    @api.model
+    @api.model_create_multi
+    @api.returns('self', lambda value: value.id)
```
pedrobaeza: I don't think you need to indicate this return decorator

hr_holidays_public/models/hr_holidays_public.py
```diff
@@ -213,11 +213,13 @@ def _update_calendar_event(self):
...
     def create(self, values):
```
pedrobaeza: Follow super argument name:
```suggestion
    def create(self, vals_list):
```

hr_holidays_public/models/hr_holidays_public.py
```diff
@@ -213,11 +213,13 @@ def _update_calendar_event(self):
...
         res = super().create(values)
-        res.meeting_id = self.env["calendar.event"].create(
-            res._prepare_holidays_meeting_values()
+        for record in res:
+            record.meeting_id = self.env["calendar.event"].create(
+            record._prepare_holidays_meeting_values()
```
pedrobaeza: Bad indentation

pedrobaeza (changes requested):

hr_holidays_public/models/hr_holidays_public.py
gurneyalex: indeed... leftover from old times :)

pedrobaeza (approved):

pedrobaeza:
We can fast-track this easy one:

/ocabot merge patch

PR #47, opened 2022-06-16
[15.0][FIX] hr_holidays_public: wrong computation in case resources in different countries

gurneyalex:
Fix an issue in hr_holidays_public where
resource.calendar::_attendance_intervals_batch_exclude_public_holidays
would return a wrong result when called with multiple resources working
in different countries.

The original implementation uses a context key employee_id to find the
employee, which only allows this method to be called with a single
employee. The fix searches for the employees related to the resources
passed in arguments, and default to the context key if there are no
provided resources.

Also rework the implementation to use a python set rather than a list
for faster in test.

### DIFF
--- a/hr_holidays_public/models/resource_calendar.py
+++ b/hr_holidays_public/models/resource_calendar.py
@@ -15,20 +15,46 @@ class ResourceCalendar(models.Model):
     def _attendance_intervals_batch_exclude_public_holidays(
         self, start_dt, end_dt, intervals, resources, tz
     ):
-        list_by_dates = (
-            self.env["hr.holidays.public"]
-            .get_holidays_list(
-                start_dt=start_dt.date(),
-                end_dt=end_dt.date(),
-                employee_id=self.env.context.get("employee_id", False),
+        holidays_by_country = {
+            False: set(
+                self.env["hr.holidays.public"]
+                .get_holidays_list(
+                    start_dt=start_dt.date(),
+                    end_dt=end_dt.date(),
+                    employee_id=self.env.context.get("employee_id", False),
+                )
+                .mapped("date")
             )
-            .mapped("date")
-        )
+        }
+        if resources:
+            employees = self.env["hr.employee.public"].search(
+                [("resource_id", "in", resources.ids)]
+            )
+            resource_country = {}
+            for employee in employees:
+                country = employee.address_id.country_id
+                resource_country[employee.resource_id.id] = country.id
+                if country.id not in holidays_by_country:
+                    holidays_by_country[country.id] = set(
+                        self.env["hr.holidays.public"]
+                        .get_holidays_list(
+                            start_dt=start_dt.date(),
+                            end_dt=end_dt.date(),
+                            employee_id=employee.id,
+                        )
+                        .mapped("date")
+                    )
+        # even if employees and resource_country are empty
+        # we still process holidays, so provide defaults
         for resource in resources:
-            interval_resource = intervals[resource.id]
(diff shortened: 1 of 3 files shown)


pedrobaeza:
Can you add a regression test that exercises the problem you have found?

gurneyalex:
@pedrobaeza currently the test is in a customer-specific module, at a higher level. I'll addapt it for this module soon but I need to hotfix this in prod first ;)

hr_holidays_public/models/resource_calendar.py
```diff
@@ -15,20 +15,42 @@ class ResourceCalendar(models.Model):
...
-                end_dt=end_dt.date(),
-                employee_id=self.env.context.get("employee_id", False),
+        holidays_by_country = {
+            False: set(
+                self.env["hr.holidays.public"]
+                .get_holidays_list(
+                    start_dt=start_dt.date(),
+                    end_dt=end_dt.date(),
+                    employee_id=self.env.context.get("employee_id", False),
+                )
```
ivantodorovich, line 25: ~In case `self.env.context.get("employee_id")` is set, shouldn't the this dict key start with the employee's country.id rather than `False` ?~

Nevermind I think I got it now

SilvioC2C (approved):

hr_holidays_public/models/resource_calendar.py
```diff
@@ -15,20 +15,42 @@ class ResourceCalendar(models.Model):
...
+                .mapped("date")
             )
-            .mapped("date")
-        )
+        }
+        if resources:
+            employees = self.env["hr.employee.public"].search(
```
yankinmax, line 31: hello @gurneyalex 
on one of projects we encounter empty recordset for `employees`, which can cause further mapping issues (KeyError) looking for the country
My suggestion is to improve little bit `resource in resources` loop:
```
for resource in resources:
    interval_resource = intervals.get(resource.id)
    attendances = []
    country = resource_country.get(resource.id, self.env["res.country"])
    holidays = holidays_by_country.get(country)
    for attendance in interval_resource._items:
        if holidays and attendance[0].date() not in holidays:
            attendances.append(attendance)
    intervals[resource.id] = Intervals(attendances)
```
This is checked on actual database

hr_holidays_public/models/resource_calendar.py
```diff
@@ -15,20 +15,46 @@ class ResourceCalendar(models.Model):
...
+                [("resource_id", "in", resources.ids)]
+            )
+            resource_country = {}
+            for employee in employees:
+                country = employee.address_id.country_id
+                resource_country[employee.resource_id.id] = country.id
+                if country.id not in holidays_by_country:
+                    holidays_by_country[country.id] = set(
```
yankinmax, line 39: to be more clear about my PR https://github.com/gurneyalex/hr-holidays/pull/2
holidays will be assigned as a set of public holidays or empty set on line 55.

yankinmax (approved):

pedrobaeza:
@gurneyalex please amend the PR without merge commits and add the regression test.

gurneyalex:
@pedrobaeza done :-)

pedrobaeza:
Thanks. I still see two commits (one being fixup). Please squash them, as the bot can't do it.

AaronHForgeFlow (approved):
Code LG

pedrobaeza:
@gurneyalex can you please squash your 2 commits into one? ocabot doesn't do it.

dreispt:
Superseded by https://github.com/OCA/hr-holidays/pull/98

PR #48, opened 2022-07-15
[13.0][FIX] Public vacations cannot be saved if variable_date is False

### LINKED ISSUE #45: 15.0 hr_holidays_public: cannot save line with variable_date = false
On 15.0 at least, it is not possible to create public holidays with variable date = false.

## Module
hr_holidays_public

## Describe the bug
On a fresh instance, when recording public holidays for a new country, saving will fail if some lines have "variable_date = False"

## To Reproduce
**Affected versions**: 15.0 at least

1. Create a new public holidays calendar for 2022 and country = France
2. Add a line with date = 2022-12-25, name = Christmas, and uncheck "variable date"
3. Click on Save

You will get a Validation Error with the following message:

> 
> The operation cannot be completed:
> - Create/update: a mandatory field is not set.
> - Delete: another model requires the record being deleted. If possible, archive it instead.
> 
> Model: Public Holidays Lines (hr.holidays.public.line), Field: Date **(date)**

image

**Expected behavior**
Saving should work

**Additional context**

Problem comes from the view, when unchecking the variable date boolean field, the date field becomes readonly and then the web client will not send the data to the backend resulting in the error when trying to create the hr.holidays.public.line record. 

Workaround is to save with the box checked, then edit, and uncheck the lines which are not variable date, but this is a really bad user experience.

cesar-tecnativa:
@Tecnativa TT37982

### DIFF
--- a/hr_holidays_public/views/hr_holidays_public_view.xml
+++ b/hr_holidays_public/views/hr_holidays_public_view.xml
@@ -30,6 +30,7 @@
                             <field
                                 name="date"
                                 attrs="{'readonly':[['variable_date', '=', False]]}"
+                                force_save="1"
                             />
                             <field name="name" />
                             <field


cesar-tecnativa:
@pedrobaeza please review

pedrobaeza:
Steps to reproduce the problem (on the PR message and the commit message)?

FernandoRomera:
In https://github.com/OCA/hr-holidays/issues/45#issue-1270315456, are the steps to reproduce the problem.

pedrobaeza:
Ok, but it should be put here and in the commit message.

cesar-tecnativa:
On a fresh instance, when recording public holidays for a new country, saving will fail if some lines have "variable_date = False"

pedrobaeza (approved):

pedrobaeza:
Let's fast-track it for fixing the problem:

/ocabot merge patch

pedrobaeza:
Please fw-port it to v14 and v15

pedrobaeza:
Merging manually as the bot is clearly not working.

@sbidoul any hint?

pedrobaeza:
Discovered the problem. Fixed in https://github.com/OCA/hr-holidays/commit/fa14cbbb7cfc76add13a8a8786b60ebd93c7c926#diff-ac5eead3f3ce4863c524fff031a87b7aecccb4a0493df087a4e1c704a1505036R18

But don't know why the display was putting it in green.

PR #51, opened 2022-07-27
[15.0][FIX] Public vacations cannot be saved if variable_date is False

### LINKED ISSUE #45: 15.0 hr_holidays_public: cannot save line with variable_date = false

cesar-tecnativa:
In https://github.com/OCA/hr-holidays/issues/45#issue-1270315456, are the steps to reproduce the problem.

### DIFF
--- a/hr_holidays_public/views/hr_holidays_public_view.xml
+++ b/hr_holidays_public/views/hr_holidays_public_view.xml
@@ -30,6 +30,7 @@
                             <field
                                 name="date"
                                 attrs="{'readonly':[['variable_date', '=', False]]}"
+                                force_save="1"
                             />
                             <field name="name" />
                             <field


pedrobaeza (approved):

pedrobaeza:
Fast-tracking this:

/ocabot merge minor

PR #64, opened 2023-02-27
[15.0] Add module hr_holidays_calendar_event_privacy

grindtildeath:
This module allows to redefine the privacy level of calendar events created from Time off requests.

### DIFF
--- a/hr_holidays_calendar_event_privacy/models/hr_leave.py
+++ b/hr_holidays_calendar_event_privacy/models/hr_leave.py
@@ -1,0 +1,18 @@
+# Copyright 2023 Camptocamp SA
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl)
+from odoo import models
+from odoo.tools import groupby
+
+
+class HrLeave(models.Model):
+    _inherit = "hr.leave"
+
+    def _prepare_holidays_meeting_values(self):
+        """Set calendar events privacy for hr leaves according to type"""
+        res = super()._prepare_holidays_meeting_values()
+        for user, holidays in groupby(self, lambda h: h.user_id):
+            for holiday, meeting_vals in zip(holidays, res.get(user.id)):
+                meeting_vals[
+                    "privacy"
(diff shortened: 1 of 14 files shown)


hr_holidays_calendar_event_privacy/models/hr_leave.py
```diff
@@ -0,0 +1,18 @@
...
+
+
+class HrLeave(models.Model):
+    _inherit = "hr.leave"
+
+    def _prepare_holidays_meeting_values(self):
+        """Set calendar events privacy for hr leaves according to type"""
+        res = super()._prepare_holidays_meeting_values()
+        for user, holidays in groupby(self, lambda h: h.user_id):
+            for holiday, meeting_vals in zip(holidays, res.get(user.id)):
```
TDu, line 14: Could there be an issue with the order the leaves are being processed in the parent function and the order they are processed here (especially because of the groupby ?
Would be good if they had a specific function in the core to set the meeting values.

TDu (commented):
also the `__init__.py` file in the test directory  is empty !

hr_holidays_calendar_event_privacy/models/hr_leave.py
grindtildeath: As both the `odoo.tools.groupby` function (https://github.com/odoo/odoo/blob/15.0/odoo/tools/misc.py#L1176)  and the `_prepare_holidays_meeting_values` use a list inside a defaultdict https://github.com/odoo/odoo/blob/15.0/addons/hr_holidays/models/hr_leave.py#L1034, we can safely assume the order of items in the list will match the order in the recordset.
Of course, it would be nicer to have a hook in Odoo, but you know :shrug:

grindtildeath:
@OCA/human-resources-maintainers can we please reopen this one? :pray:

nimarosa:
@grindtildeath ready.

petrus-v (approved):
Code LGTM, thanks !

grindtildeath:
@dreispt CI must be fixed now :slightly_smiling_face:

PR #67, opened 2023-03-03
[ADD] hr_holidays_auto_extend

etobella:
With this module, we will be able to extend automatically a leave when the period is reached.

This might be necessary for leaves without a clear return date for the employee.


kluna1998 (approved):
LGTM 👍🏽

luisDIXMIT (approved):
LGTM!

PR #70, opened 2023-04-11
[15.0][ADD] hr_holidays_leave_self_approve

florentx:


rafaelbn:
I think an better explanation should be in README as this can be done in Odoo Community without any module by:

- Setting approval to "no validation" or (better)
- Just setting "By Employee's Approver" and changing the domain of the field in the employee to it self.

Just curious, I would not use this module even when I allow my employees self-allow approvals 😄 

Not blocking anything! Regards

florentx:
> I think an better explanation should be in README as this can be done in Odoo Community without any module by:
> 
>     * Setting approval to "no validation" or (better)
> 
>     * Just setting "By Employee's Approver" and changing the domain of the field in the employee to it self.
> 

This add-on is useful for our use case because:

1. They want a validation, even if it is a self validation
2. Changing the domain of the field to the employee itself does not work because of this check in standard (Odoo 15)
```
# File "addons/hr_holidays/models/hr_leave.py"

                    if holiday.employee_id == current_employee:
                        raise UserError(_('Only a Time Off Manager can approve/refuse its own requests.'))
```

hr_holidays_leave_self_approve/__manifest__.py
```diff
@@ -0,0 +1,15 @@
+# Copyright 2023 Camptocamp SA
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
+
+{
+    "name": "Self Approve Leaves",
+    "version": "15.0.1.0.0",
+    "license": "AGPL-3",
+    "summary": "Allow employee to approve his own leaves",
+    "author": "Camptocamp SA, Odoo Community Association (OCA)",
```
simahawk:
```suggestion
    "author": "Camptocamp, Odoo Community Association (OCA)",
```

hr_holidays_leave_self_approve/tests/test_hr_holidays_leave_self_approve.py
```diff
@@ -0,0 +1,79 @@
+# Copyright 2023 Camptocamp SA
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
+
+from datetime import datetime, timedelta
+
+from odoo.tests.common import TransactionCase
+
+
+class TestHolidaysAutoValidate(TransactionCase):
+    def setUp(self):
```
simahawk: pls use setUpClass and disable tracking

hr_holidays_leave_self_approve/__manifest__.py
```diff
@@ -0,0 +1,15 @@
+# Copyright 2023 Camptocamp SA
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
+
+{
+    "name": "Self Approve Leaves",
+    "version": "15.0.1.0.0",
+    "license": "AGPL-3",
+    "summary": "Allow employee to approve his own leaves",
```
simahawk:
```suggestion
    "summary": "Allow employees to approve their own leaves when allowed",
```

simahawk (changes requested):

simahawk (approved):

gurneyalex (approved):

PR #71, opened 2023-04-19
[15.0][FIX] hr_holidays_natural_period: Avoid error when saving leave

victoralmau:
Avoid error when saving leave.

@Tecnativa TT42608

### DIFF
--- a/hr_holidays_natural_period/models/hr_leave_type.py
+++ b/hr_holidays_natural_period/models/hr_leave_type.py
@@ -1,4 +1,4 @@
-# Copyright 2020 Tecnativa - Víctor Martínez
+# Copyright 2020-2023 Tecnativa - Víctor Martínez
 # License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
 
 from odoo import fields, models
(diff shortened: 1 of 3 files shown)


chienandalu (approved):
Thanks :)

hr_holidays_natural_period/models/hr_leave_type.py
```diff
@@ -11,3 +11,23 @@ class HrLeaveType(models.Model):
...
+    def _get_employees_days_per_allocation(self, employee_ids, date=None):
+        """We need to set request_unit as 'day' to avoid the calculations being done
+        as hours.
+        Related code:
+        hr_holidays/models/hr_leave_type.py#L313
+        hr_holidays/models/hr_leave_type.py#L367
+        """
+        old_request_unit_data = {}
+        for item in self:
+            old_request_unit_data[item.id] = item.request_unit
```
pedrobaeza, line 24: Don't store (and change) all the types, just those with natural days.

hr_holidays_natural_period/__manifest__.py
```diff
@@ -12,5 +12,5 @@
     "installable": True,
     "depends": ["hr_holidays"],
     "maintainers": ["victoralmau"],
-    "demo": ["demo/hr_leave_type_data.xml"],
+    "demo": ["demo/hr_leave_type_demo.xml"],
```
pedrobaeza: This change requires a migration script to try to delete the type.

victoralmau: I have left the file name as it was.

pedrobaeza: No, it's correct to be demo data, as it was an error to put it as data, but it should delete old occurrences through delete_safely... method in openupgradelib.

victoralmau: I'm sorry but I don't understand.

They have always been defined as demo data, now we are just renaming the file to make it more consistent.

pedrobaeza: Ah, OK, my fault. Sorry, I saw data > demo, and I think this is changing the other part (the key, changing its meaning). It's OK then to just rename and nothing more.

pedrobaeza (approved):
/ocabot merge patch

PR #80, opened 2023-07-04
[15.0][mig] hr_holidays_credit: Migration to 15.0

dsolanki-initos:


RFoerster:
Hi - Thanks for taking the time to migrate this module! I was wondering why - in your last Commit (https://github.com/initOS/hr-holidays/commit/c426bf732882a9a95f27748c57783b9b44d9727a) you've changed the OR-Condition to an AND-Condition for the hr_leave-Domain. In my opinion this is wrong as now "Requires Allocation" has to be "no" when "Allow Credit" is "True"? Thanks

PR #81, opened 2023-07-04
[15.0][FIX] hr_holidays_public: allow to search holidays to non hr users

AaronHForgeFlow:
When users with no privileges run the method get_holidays_list an access error is raised. This won't happen with only this module installed but on modules built on top of this one this is possible and the fix won't hurt IMHO.

@ForgeFlow


pedrobaeza:
The sudo should be done on the modules where is the problem, not here IMO.

AaronHForgeFlow:
I just don't want to overwrite the method just to add a sudo there. But I am afraid I don't have more options.

PR #83, opened 2023-07-05
[15.0][ADD] hr_holidays_public_city: New module

victoralmau:
Add Related Cities to public holidays.
It is now possible to filter by city (optional).

@Tecnativa TT43035


hr_holidays_public/__manifest__.py
```diff
@@ -14,7 +14,7 @@
     "Odoo Community Association (OCA),",
     "summary": "Manage Public Holidays",
     "website": "https://github.com/OCA/hr-holidays",
-    "depends": ["hr_holidays"],
+    "depends": ["hr_holidays", "base_location"],
```
chienandalu: Maybe not everybody is happy to pull this new dependency...

victoralmau: Totally agree, but then I think it would be necessary to add a new module with both dependencies (`hr_holidays_public` + `base_location`).

The model `res.city` is added with `base_address_city`
`city_id` field in `res.company` is added with `base_location`

chienandalu: Yes... what do you think about it @pedrobaeza ?

pedrobaeza: AFAIK, you don't need res.company->city_id. You need res.partner->city_id. So the new dependency is `base_address_city`, not `base_location`. Adding such dependency on a stable module? Well, it can be debatable, but it's not a big deal IMO, as it's a core Odoo module that is also very transparent once installed.

Let's see what other contributors like @ivantodorovich or @petrus-v think about it.

ivantodorovich: I would prefer to have a glue module for this, rather than adding the new dependency, even if it's `base_address_city`

victoralmau: Now the dependency added is `base_address_city`, so do you think this functionality should go in a new module?

petrus-v: I'm not using `base_address_city` that's much and I'm learning that there are cities that gives more days off (I'm wondering which ones ? I'm so curious to know, where should I work to get more days off, we could open a postal box for remote employes there :p).

So technically spoken I would prefer a new module that depends on both `hr_holidays_public` + `base_address_city` to avoid change on res partner view and maintains code that I do not really needs.

In an other way as long `base_address_city` appears to be transparent for end user and manage case where city_id is not used it's kind of ok with probably a roadmap to suggest to refactor the module ! but I would say it's pity to not do it first !

my  2cts

nimarosa: For me i think it's not a big deal to add this dependency and i think the improvement is a good one. So in my opinion i'm okay with this change, and also will like to see it ported to 14.0 version since it seems very useful for some use cases.

nimarosa (approved):

hr_holidays_public/__manifest__.py
chienandalu: >  that's much and I'm learning that there are cities that gives more days off

I don't know of other countries, but in Spain there are days allocated by the nation, days allocated by the regions/states and days allocated by the municipalities; so every spanish has the same holidays but they can be on different days depending on where they live.

petrus-v: > > that's much and I'm learning that there are cities that gives more days off
> 
> I don't know of other countries, but in Spain there are days allocated by the nation, days allocated by the regions/states and days allocated by the municipalities; so every spanish has the same holidays but they can be on different days depending on where they live.

ok good to know ! where employee lives or according the office place ? (As far I've understood source code it's get from the employee office place !)

nimarosa: @petrus-v Not always. The business can have people working in a lot of different cities in the country, some remote, some on-site. So sometimes is according to where the employee lives. Also, an employee who works in a office, probably the office (the one he need to attend to) is in the same city so office city and employee city usually to match.

victoralmau:
Finally then I make these changes in a glue module, ok?

victoralmau:
New module proposed

ivantodorovich (approved):
Thanks a lot for taking care of this ! 
It looks good !

petrus-v (approved):
Thanks for the split !

code review LGTM

victoralmau:
This is ready to merge right?

hr_holidays_public/models/hr_holidays_public.py
```diff
@@ -55,6 +55,25 @@ def name_get(self):
             result.append((rec.id, rec.display_name))
         return result
 
+    def _prepare_holidays_list_states_filter_domain(
```
pedrobaeza: `prepare` is for dictionary vals return. Use `_get_domain_` or similar.

hr_holidays_public_city/models/hr_holidays_public.py
```diff
@@ -0,0 +1,66 @@
...
+
+    def _prepare_domain_check_date_state_one(self):
+        domain = super()._prepare_domain_check_date_state_one()
+        domain += [("city_ids", "=", False)]
+        return domain
+
+    def _prepare_holidays_meeting_values(self):
+        res = super()._prepare_holidays_meeting_values()
+        if self.city_ids:
+            res["description"] += ": " + ", ".join(self.city_ids.mapped("name"))
```
pedrobaeza, line 63: It should put the corresponding city of the employee, not the whole city list.

victoralmau: Currently this is not possible because the `calendar.event` record is created for each `hr.holidays.public.line` (https://github.com/OCA/hr-holidays/blob/15.0/hr_holidays_public/models/hr_holidays_public.py#L221).

pedrobaeza: OK, my fault. I was thinking this was the specific `hr.leave` corresponding entry.

hr_holidays_public/models/hr_holidays_public.py
```diff
@@ -147,6 +156,21 @@ def _check_date_state(self):
         for line in self:
             line._check_date_state_one()
 
+    def _prepare_domain_check_date_state_one_state_ids(self):
```
pedrobaeza: These methods have the same problem on the naming. Please remember always to apply the same comment to the whole diff.

hr_holidays_public/models/hr_leave.py
```diff
@@ -48,9 +48,7 @@ def _compute_number_of_hours_display(self):
             super(HrLeave, leave)._compute_number_of_hours_display()
         return super(HrLeave, self - to_serialize)._compute_number_of_hours_display()
 
-    @api.model
-    def get_unusual_days(self, date_from, date_to=None):
-        res = super().get_unusual_days(date_from, date_to=date_to)
+    def _prepare_domain_from_get_unusual_days(self, date_from, date_to=None):
```
pedrobaeza: This one is missing.

hr_holidays_public_city/models/hr_leave.py
```diff
@@ -0,0 +1,25 @@
...
+        )
+        # Add city domain
+        city_id = self.env.user.employee_id.address_id.city_id.id
+        if not city_id:
+            city_id = self.env.company.partner_id.city_id.id or False
+        if city_id:
+            domain.extend(
+                [
+                    "|",
+                    ("city_ids", "in", [city_id]),
```
pedrobaeza: Why not?

```suggestion
                    ("city_ids", "=", city_id),
```

pedrobaeza (approved):
/ocabot merge patch

PR #85, opened 2023-07-25
[16.0][IMP] hr_holidays_public: Allow inheritance by other modules

victoralmau:
FWP from 15.0: https://github.com/OCA/hr-holidays/pull/83

Allow inheritance by other modules.

@Tecnativa TT43035

### DIFF
--- a/hr_holidays_public/models/hr_holidays_public.py
+++ b/hr_holidays_public/models/hr_holidays_public.py
@@ -55,6 +55,23 @@ class HrHolidaysPublic(models.Model):
             result.append((rec.id, rec.display_name))
         return result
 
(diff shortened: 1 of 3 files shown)


chienandalu (approved):

pedrobaeza (approved):
/ocabot merge patch

SkiBY:
Sorry, but after some tests, it seems, module doesn't work correctly with leave days counting in 16 and duplicates Public Holidays based on resource.calendar.leaves

shouldn't be this added to readme instedad of v.12 notes?

victoralmau:
Please create an issue explaining all the steps to configure it the same, what is the result obtained and what should be the result.

Thank you.

PR #88, opened 2023-09-21
Migration to 16.0

rooniweb:
Module `hr_holidays_leave_auto_approve` Migration :

- Moved auto_approve_policy field from `validity` to  `leave_validation` in hr_holidays_status.xml.
- Alignment with v16 changes.

This ensures module compatibility with the new version.


pedrobaeza:
Thanks for the contribution.

Please preserve commit history following technical method explained in https://github.com/OCA/maintainer-tools/wiki/Migration-to-version-16.0.

If the jump is between several versions, you have to modify the source branch in the main command to accommodate it to this circumstance.

PR #91, opened 2023-10-04
[ADD][14.0] New module hr_holidays_validator_list

victorvermot:
A module making replacing the current leave_manager_id field by a many2many field called leave_manager_ids. Thus allowing a user to have several leave managers.

TODO: 
- [ ] More unit tests


hr_holidays_validator_list/models/hr_employee.py
```diff
@@ -0,0 +1,47 @@
...
+        "res.users",
+        string="Time Off Managers",
+        store=True,
+        readonly=False,
+        help="""Select the users responsible for approving 'Time Off' of this employee.\n
+        If empty, the approval is done by an Administrator
+        or Approver (determined in settings/users).""",
+    )
+
+    def _set_leave_manager_id_from_leave_manager_ids(self, values):
```
StephaneMangin: You should have used a stored readable only computed method on leave_manager_id with a depends on leave_manager_ids to propagate any change.

hr_holidays_validator_list/models/hr_employee.py
```diff
@@ -0,0 +1,47 @@
...
+        if "leave_manager_ids" in values:
+            if not len(values["leave_manager_ids"][0][-1]):
+                values["leave_manager_id"] = None
+            elif "leave_manager_ids" in values.keys():
+                values["leave_manager_id"] = values["leave_manager_ids"][0][-1][-1]
+        return values
+
+    def _add_leave_manager_ids_in_group(self, values):
```
StephaneMangin, line 29: An onchange would have suffice to affect groups on managers.

hr_holidays_validator_list/models/hr_employee.py
victorvermot: You're right I am going to check that

hr_holidays_validator_list/models/hr_employee.py
```diff
@@ -0,0 +1,47 @@
...
+
+class HrEmployeeBase(models.AbstractModel):
+    _inherit = "hr.employee.base"
+
+    leave_manager_ids = fields.Many2many(
+        "res.users",
+        string="Time Off Managers",
+        store=True,
+        readonly=False,
+        help="""Select the users responsible for approving 'Time Off' of this employee.\n
```
yvaucher: This double the end line as `"""` will preserve the return line and the `\n` is added. Is it intended?

hr_holidays_validator_list/models/hr_leave_allocation.py
```diff
@@ -0,0 +1,32 @@
...
+from odoo import models
+
+
+class HolidaysAllocation(models.Model):
+    """Allocation Requests Access specifications: similar to leave requests"""
+
+    _inherit = "hr.leave.allocation"
+
+    def activity_update(self):
+        """uUpdates activity for all leave_manager_ids"""
```
yvaucher:
```suggestion
        """Updates activity for all leave_manager_ids"""
```

hr_holidays_validator_list/models/hr_employee.py
yvaucher: I disagree, we are on hr_employee, an onchange should only affect the values of the current form. Groups won't be available from here.

hr_holidays_validator_list/models/hr_employee.py
yvaucher: @StephaneMangin The field `leave_manager_ids` is already a computed stored field. Question is can we insert nicely the set of the values from the M2M nicely or not. We might want to keep leave_manager_ids value to it's current natural computation.

@victorvermot 

What we could probably do to improve the code is:

Override https://github.com/odoo/odoo/blob/17.0/addons/hr_holidays/models/hr_employee.py#L152-L159
this way:

``` python
    @api.depends('parent_id', 'leave_manager_ids')
    def _compute_leave_manager(self):
            
        for employee in self:
            if self.leave_manager_ids:
               self.leave_manager_id = leave_manager_ids[0]     
           else:
                super()._compute_leave_manager()
```

yvaucher (changes requested):
Small typos and a proposal of improvement of the code.

yvaucher (approved):

cyrilmanuel (approved):
lgtm

gurneyalex:
@victorvermot you should check the CI: the tests are failing.

PR #98, opened 2023-11-18
[FIX] hr_holidays_public: wrong computation in case resources in different countries

dreispt:
Fix an issue in hr_holidays_public where
resource.calendar::_attendance_intervals_batch_exclude_public_holidays
would return a wrong result when called with multiple resources working
in different countries.

The original implementation uses a context key employee_id to find the
employee, which only allows this method to be called with a single
employee. The fix searches for the employees related to the resources
passed in arguments, and default to the context key if there are no
provided resources.

Also rework the implementation to use a python set rather than a list
for faster in test.


dreispt:
Supersedes https://github.com/OCA/hr-holidays/pull/47

pedrobaeza:
Is this happening in 16?

PR #100, opened 2023-11-29
[FIX] hr_holidays_public: Don't populate partner_ids on calendar event

moctarjallo:
When you configure a public holiday in `hr.holidays.public.line` then a `calendar.event` calendar event linked to this public holiday (`meeting_id` field in `hr.holidays.public.line`) is automatically created with the `partner_ids` field filled in by the partner of the user who created the holiday.

This PR modifies so that when creating the calendar event `calendar.event` corresponding to the configured holiday `hr.holidays.public.line`, the `partner_ids` field is not filled in (leave empty).

### DIFF
--- a/hr_holidays_public/models/hr_holidays_public.py
+++ b/hr_holidays_public/models/hr_holidays_public.py
@@ -217,6 +217,7 @@ class HrHolidaysPublicLine(models.Model):
             "user_id": SUPERUSER_ID,
             "privacy": "confidential",
             "show_as": "busy",
+            "partner_ids": [],
         }
         if categ_id:
             meeting_values.update({"categ_ids": [(6, 0, categ_id.ids)]})


hr_holidays_public/models/calendar_event.py
```diff
@@ -0,0 +1,13 @@
+# Copyright 2023 XCG-CONSULTING - Moctar Diallo <moctarjallo@gmail.com>
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
+
+from odoo import models
+
+
+class CalendarEvent(models.Model):
+    _inherit = "calendar.event"
+
+    def default_get(self, fields_list):
```
abhinavvaidya-tsd: Please add @api.model decorator for better Odoo convention.

abhinavvaidya-tsd (approved):

hr_holidays_public/__manifest__.py
```diff
@@ -4,7 +4,7 @@
 
 {
     "name": "HR Holidays Public",
-    "version": "16.0.1.1.2",
+    "version": "16.0.1.2.0",
```
pedrobaeza: Please don't change manually the version, It's done by the bot on merge. This way, we avoid possible conflicts if there are several PRs for the same module done.

hr_holidays_public/models/calendar_event.py
```diff
@@ -0,0 +1,14 @@
...
+
+from odoo import api, models
+
+
+class CalendarEvent(models.Model):
+    _inherit = "calendar.event"
+
+    @api.model
+    def default_get(self, fields_list):
+        defaults = super().default_get(fields_list)
```
pedrobaeza: Please don't do it this way, as this is affecting all the calendar events. Just pass `"partner_ids": False` in the dictionary creation instead.

pedrobaeza (changes requested):

hr_holidays_public/models/hr_holidays_public.py
```diff
@@ -217,6 +217,7 @@ def _prepare_holidays_meeting_values(self):
             "user_id": SUPERUSER_ID,
             "privacy": "confidential",
             "show_as": "busy",
+            "partner_ids": False,
```
pedrobaeza: It seems this should be:

```suggestion
            "partner_ids": [],
```

pedrobaeza (commented):
And please follow commit message guidelines:

https://github.com/OCA/odoo-community.org/blob/master/website/Contribution/CONTRIBUTING.rst#71commit-message

hr_holidays_public/models/hr_holidays_public.py
moctarjallo: Right!

pedrobaeza:
The only thing left is the commit message. Please put for example:

```
[FIX] hr_holidays_public: Don't populate partner_ids on calendar event
```

pedrobaeza (approved):
/ocabot merge patch

Please fw-port it to v17

PR #102, opened 2023-12-14
[17.0][MIG] hr_holidays_public: Migration to 17.0

nguyenminhchien:


gurneyalex (approved):

rafaelbn:
Hello @gurneyalex @nguyenminhchien  , could I ask please when are we going to use in OCA de Odoo core public holidays? We are migrating since Odoo 15 something is in Odoo core 🤔 .

Thank you very much! 😄 See you soon! ❤️

PR #106, opened 2024-02-05
test hr_holidays_calendar_event_privacy separately

gurneyalex:
The tests on the 15.0 branch are broken because of a failing test in hr_holidays_calendar_event_privacy. This looks like a bad interaction with another module so I'm moving the testing of hr_holidays_calendar_event_privacy to a separate batch. 

Done as part of an update of the copier template

### DIFF
--- a/.copier-answers.yml
+++ b/.copier-answers.yml
@@ -1,7 +1,8 @@
 # Do NOT update manually; changes here will be overwritten by Copier
-_commit: v1.17.2
+_commit: v1.20
 _src_path: gh:oca/oca-addons-repo-template
 ci: GitHub
+convert_readme_fragments_to_markdown: false
 generate_requirements_txt: true
 github_check_license: true
 github_ci_extra_env: {}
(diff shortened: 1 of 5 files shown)


pedrobaeza:
Duplicate the builds is not good. What is the error and can't it be prevented other way (for example, just detecting if `config["test_enable"]` and without a specific context for disabling the incompatible behavior if not testing the own module)?

gurneyalex:
@pedrobaeza the error on the 15.0 branch is https://github.com/OCA/hr-holidays/actions/runs/7276826929/job/19827470569#step:8:37 which is different from the one I get on this build. 

I'll check with @grindtildeath about the failure

gurneyalex:
The tests were green when the last merge on the branch was made on Dec 9, 2023, and started failing on a translation commit on Dec. 20, 2023. So the cause if the red CI is not on the module I'm splitting, but likely in a dependency or in an change in Odoo :-( 

I'm closing this PR, and will dig a bit deeper later.

PR #110, opened 2024-02-20
[16.0] hr_holidays_public: Deprecate use of employee_id in favour of partner_id

grindtildeath:
Helper functions to find public holidays did allow to use an hr.employee to filter country and states based on the employee address.

Since only the address of the employee was used, modifying the functions to use a res.partner instead of an hr.employee allows more possibilities such as checking public holidays for customers and suppliers.

### DIFF
--- a/hr_holidays_public/models/hr_holidays_public.py
+++ b/hr_holidays_public/models/hr_holidays_public.py
@@ -3,10 +3,13 @@
 # License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
 
 import datetime
+import logging
 from datetime import date
 
 from odoo import SUPERUSER_ID, _, api, fields, models
 from odoo.exceptions import ValidationError
+
+_logger = logging.getLogger(__name__)
 
 
 class HrHolidaysPublic(models.Model):
@@ -55,16 +58,16 @@ class HrHolidaysPublic(models.Model):
             result.append((rec.id, rec.display_name))
         return result
 
-    def _get_domain_states_filter(self, pholidays, start_dt, end_dt, employee_id=None):
-        employee = False
-        if employee_id:
-            employee = self.env["hr.employee"].browse(employee_id)
+    def _get_domain_states_filter(
+        self, pholidays, start_dt, end_dt, employee_id=None, partner_id=None
+    ):
+        partner = self._get_partner_deprecated_employee(partner_id, employee_id)
         states_filter = [("year_id", "in", pholidays.ids)]
-        if employee and employee.address_id and employee.address_id.state_id:
+        if partner and partner.state_id:
             states_filter += [
                 "|",
                 ("state_ids", "=", False),
-                ("state_ids", "=", employee.address_id.state_id.id),
+                ("state_ids", "=", partner.state_id.id),
             ]
         else:
             states_filter.append(("state_ids", "=", False))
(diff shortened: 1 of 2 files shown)


hr_holidays_public/models/hr_holidays_public.py
```diff
@@ -86,50 +90,69 @@ def get_holidays_list(
...
+        if employee_id is not None:
+            _logger.warning(
+                "Use of employee_id for hr.public.holidays is deprecated. "
+                "Please use partner_id instead."
+            )
+            employee = self.env["hr.employee"].browse(employee_id)
+            partner = employee.address_id
+        if partner_id:
+            if partner:
+                _logger.error(
```
simahawk:
```suggestion
                _logger.warning(
```

hr_holidays_public/models/hr_holidays_public.py
```diff
@@ -86,50 +90,69 @@ def get_holidays_list(
...
             hol_date = holidays_lines.filtered(lambda r: r.date == selected_date)
             if hol_date.ids:
                 return True
         return False
 
+    def _get_partner_deprecated_employee(self, partner_id, employee_id):
+        # TODO: Drop function in next migration
+        employee = False
+        partner = False
+        if employee_id is not None:
```
mmequignon, line 140: what if you call this method with `(False, False)` ?

grindtildeath: From where it's called, it's supposed to have a value or be None, so no need to handle such a case

jbaudoux (approved):

simahawk (approved):

pedrobaeza (commented):
Can you please explain the rationale behind this change?

jbaudoux:
> Can you please explain the rationale behind this change?

We need this for planning the shipments to the customers. Some (country) state have specific bank holidays and we need to plan transport accordingly

cyrilmanuel (approved):

cyrilmanuel:
Hello @pedrobaeza little ping to know if it can be merge :)

hr_holidays_public/models/hr_holidays_public.py
```diff
@@ -55,16 +58,17 @@ def name_get(self):
...
 
-    def _get_domain_states_filter(self, pholidays, start_dt, end_dt, employee_id=None):
-        employee = False
-        if employee_id:
-            employee = self.env["hr.employee"].browse(employee_id)
+    def _get_domain_states_filter(
+        self, pholidays, start_dt, end_dt, employee_id=None, partner_id=None
+    ):
+        partner = self._get_partner_deprecated_employee(partner_id, employee_id)
+
```
pedrobaeza: Don't add empty lines inside a method

hr_holidays_public/models/hr_holidays_public.py
```diff
@@ -86,50 +90,71 @@ def get_holidays_list(
         :param employee_id: ID of the employee
         :return: recordset of hr.holidays.public.line
         """
+        partner = self._get_partner_deprecated_employee(partner_id, employee_id)
+
```
pedrobaeza: Don't add empty lines inside a method

pedrobaeza (changes requested):
After analyzing this, it's a bit weird to have something prefixed with `hr`, but using a partner, but it's OK if you don't deprecate ever the `employee_id` argument. It may be needed because in any override, you may have some information on the employee that vary the result. Losing the argument, you won't be able to access to this info.

grindtildeath:
@pedrobaeza I kept the possiblity to use `employee_id` on purpose to avoid any breaking change in a release version.

IMO for future versions public holidays should be taken out of the hr module as it's not necessarily linked only to hr. We could have a calendar_holidays_public module and then we could bridge the gap to the hr keeping hr_holidays_public.

grindtildeath:
@pedrobaeza Can you please update your review?

pedrobaeza (commented):
OK, let's go with the split in future version.

/ocabot merge major

pedrobaeza (approved):

PR #111, opened 2024-02-20
[17.0][FW] hr_holidays_public: Deprecate use of employee_id in favour of partner_id

grindtildeath:
Helper functions to find public holidays did allow to use an hr.employee to filter country and states based on the employee address.

Since only the address of the employee was used, modifying the functions to use a res.partner instead of an hr.employee allows more possibilities such as checking public holidays for customers and suppliers.

Forward port of:
 - [x] #110

### DIFF
--- a/hr_holidays_public/models/hr_holidays_public.py
+++ b/hr_holidays_public/models/hr_holidays_public.py
@@ -3,10 +3,13 @@
 # License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
 
 import datetime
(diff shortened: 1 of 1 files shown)


simahawk (approved):

Abranes (approved):

pedrobaeza:
Please rebase and check the CI.

grindtildeath:
@pedrobaeza done

pedrobaeza:
Still red

PR #124, opened 2024-06-05
[13.0][IMP] hr_holidays_credit: add unit in form selection widget & plurals management

maisim:
port from 12.0

https://github.com/OCA/hr/pull/1356

PR #125, opened 2024-06-05
[14.0][FW] [13.0][IMP] hr_holidays_credit: add unit in form selection widget & plurals management

maisim:
Port of #124 from 13.0 to 14.0.

### DIFF
--- a/hr_holidays_credit/models/hr_leave_type.py
+++ b/hr_holidays_credit/models/hr_leave_type.py
@@ -43,22 +43,49 @@ class HrLeaveType(models.Model):
 
             extra = None
             if record.allocation_type != "no" and context_employee_id:
-                if record.virtual_remaining_leaves >= 0:
-                    if record.allow_credit:
-                        extra = _("%g available + credit")
-                    else:
-                        extra = _("%g available")
-                    extra = extra % (
-                        float_round(record.virtual_remaining_leaves, precision_digits=2)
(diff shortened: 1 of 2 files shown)


maisim:
Maybe not directly related with this module but,

On form loading, the quantity available miss in the input:
image

it's ok in the choices:
image


and once we click on the select and choose an entry, it's ok in the input
image

An idea ?

edit : The problem is not present on 13.0

maisim:
The field definition in 13.0:
```
    holiday_status_id = fields.Many2one(
        "hr.leave.type", string="Time Off Type", required=True, readonly=True,
        states={'draft': [('readonly', False)], 'confirm': [('readonly', False)]},
        domain=[('valid', '=', True)])
```

in 14:
```
    holiday_status_id = fields.Many2one(
        "hr.leave.type", compute='_compute_from_employee_id', store=True, string="Time Off Type", required=True, readonly=False,
        states={'cancel': [('readonly', True)], 'refuse': [('readonly', True)], 'validate1': [('readonly', True)], 'validate': [('readonly', True)]},
        domain=[('valid', '=', True)])
    validation_type = fields.Selection(string='Validation Type', related='holiday_status_id.leave_validation_type', readonly=False)
    # HR data
```

I also realize that we completely replace the name_get method, which may not always be desirable, let's dig deeper.

PR #137, opened 2024-08-12
[14.0][ADD]hr_holidays_cumulative_timeoff: new module

PicchiSeba:
This new module adds a couple of fields (`cumulative_remaining_timeoff`, `cumulative_used_timeoff`) used to help leave administrators and officers understand the amount of reserved leave units at a certain point in time.


hr_holidays_cumulative_timeoff/readme/DESCRIPTION.rst
```diff
@@ -0,0 +1,2 @@
+This module adds a couple of fields showing how many time-off units (days/hours) have
+been used at the time of each time off, dividev by tome-off type.
```
HekkiMelody: **chore**: 

```suggestion
been used at the time of each time off, divided by time-off type.
```

hr_holidays_cumulative_timeoff/readme/DESCRIPTION.rst
```diff
@@ -0,0 +1,2 @@
+This module adds a couple of fields showing how many time-off units (days/hours) have
```
HekkiMelody, line 1: **chore**: Add a USAGE file, describe where we can see the report.

HekkiMelody (changes requested):
Code and functional review:

There is some weirdness when it comes to calculating the new fields.

For example, this is the result

image

for these leaves

image

Also, it'd be best if all the cells in the pivot were filled, since it's cumulative. For example, in the first screenshot, Ronnie Hart doesn't have new parental leaves in September but the pivot should show how many parental leaves he used thus far (from the beginning of time, until September, so including August)

PR #127, opened 2024-06-17
[17.0][ADD] hr_holidays_remaining_leaves

sonhd91:
The failing test not related to this module

PR #139, opened 2024-08-22
[17.0][ADD] hr_holidays_remaining_leaves

xaviedoanhduy:
The original development of this module has been done by Mint System.
It can be found in: https://github.com/Mint-System/Odoo-Apps-HR/tree/16.0/hr_holidays_remaining_leaves
This module has been ported to the OCA with their agreement, cc @janikvonrotz 

### Supersedes: https://github.com/OCA/hr-holidays/pull/127


pedrobaeza (changes requested):
- README.md file shouldn't be touched.
- The commit history contains other modules. Please squash story to reduce it and remove that noise.
- I insist the name should be singular in the extra bits. `hr_holidays_employee_remaining` should serve.

nilshamerlinck:
Hi @pedrobaeza 
- respectfully, I'm not convinced that `hr_holidays_employee_remaining` is a better name than `hr_holidays_remaining_leaves` for a module whose purpose is to "Show remaining leaves per employee in allocation overview"
- in the end of the day what matters imho is to have a good distinctive and explicit name; I think the [singular form naming rule](https://github.com/OCA/odoo-community.org/blob/master/website/Contribution/CONTRIBUTING.rst#11modules) was mainly for the odoo models (with some exceptions, to align on odoo inconsistencies), but not necessarily for the extra bits, especially if it ends up making the module's name confusing

pedrobaeza:
It was just a suggestion. What I try is to avoid the plurals, and let me tell you why the guideline about singular forms is there: this comes from the past (6.0, 6.1) where the module names where a total nightmare of mix of plurals, company names, etc, and it was very easy to misspell and commit typos while putting that names in the command line, in automation files, etc. A recurrent one is the `s` of the plural, specially when they are in the middle of the module name. That's why this rule exists. The exception is that Odoo doesn't follow these rules, and having `hr_holidays`, but extension modules being for example `hr_holiday_extension` will be weird, so we keep the plural in the prefix.

PR #140, opened 2024-09-02
[FIX] [16.0] resource_leaves_geographic: Make leave datetimes aware with user tz

Shide:
Make leave datetimes aware with user tz

In the calendar, leaves marks the previous day as a holiday when the user's time zone is greater than UTC.

MT-6858 @moduon @rafaelbn @EmilioPascual @yajo @loida-vm please review if you want :)


loida-vm (approved):
It works!
LGTM! 👍🏼  Thank you @Shide

yajo (approved):

PR #141, opened 2024-09-12
[17.0][MIG] hr_holidays_natural_period: Migration to version 17.0

carlos-lopez-tecnativa:
@Tecnativa TT50070

@pedrobaeza @victoralmau @peluko00 could you please take a look?


hr_holidays_natural_period/models/hr_leave_allocation.py
```diff
@@ -0,0 +1,12 @@
+# Copyright 2024 APSL-Nagarro - Antoni Marroig Campomar
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
+
+from odoo import fields, models
+
+
+class HrLeaveAllocation(models.Model):
+    _inherit = "hr.leave.allocation"
+
+    type_request_unit = fields.Selection(
```
pedrobaeza, line 11: Why adding this? I don't think it's needed.

carlos-lopez-tecnativa: Because in V16 it was a `related` field, but now it is `computed` and can take its value from the `hr.leave.type's` `request_unit` field. However, options from the Selection are not inherited (as occurred in V16).
See https://github.com/odoo/odoo/blob/54edf00c6ad81f8143983820ecb43501199998fe/addons/hr_holidays/models/hr_leave_allocation.py#L347

pedrobaeza: OK, please include this explanation summarized as a comment before the field declaration for future references.

carlos-lopez-tecnativa: Done

carlos-lopez-tecnativa:
The test failed after this PR  https://github.com/odoo/odoo/pull/153044.

victoralmau:
> The test failed after this PR [odoo/odoo#153044](https://github.com/odoo/odoo/pull/153044).

I'm sorry but I don't see any relation, can you rebase to check the error that occurs (in any case it will have to be solved).

carlos-lopez-tecnativa:
> > The test failed after this PR [odoo/odoo#153044](https://github.com/odoo/odoo/pull/153044).
> 
> I'm sorry but I don't see any relation, can you rebase to check the error that occurs (in any case it will have to be solved).

Because Odoo removed the field `name` from the view and now uses `private_name`, I removed it from the test since `name` is computed.
https://github.com/odoo/odoo/pull/153044/files#diff-668cc4892b604ac124476733a1f547a32a2314be62e6e038fa876e29996fe941L165

victoralmau (approved):
Code and functional review OK.

I don't understand the error in the tests indicated https://github.com/OCA/hr-holidays/pull/141#issuecomment-2352651174 (I don't see any error, PR and locally).

pedrobaeza (approved):
/ocabot merge nobump

PR #144, opened 2024-10-02
[17.0][FW] hr_holidays_public: port from 14.0

xaviedoanhduy:

PR #147, opened 2024-10-14
[18.0][MIG] hr_holidays_public: Migration to 18.0

xaviedoanhduy:
### Change on version 18.0

- `number_of_hours_display` fields has been removed in https://github.com/odoo/odoo/pull/156117
### On the occasion

- I also removed some redundant code in tests (no need to call the denpends function)

### OCA - port

-  https://github.com/OCA/hr-holidays/pull/144


mmequignon (approved):

hr_holidays_public/models/hr_holidays_public.py
```diff
@@ -108,15 +101,14 @@ def get_holidays_list(
...
+    def is_public_holiday(self, selected_date, partner_id=None):
         """
         Returns True if selected_date is a public holiday for the employee
         :param selected_date: datetime object
-        :param employee_id: ID of the employee
         :param partner_id: ID of the partner
         :return: bool
         """
-        partner = self._get_partner_deprecated_employee(partner_id, employee_id)
+        partner = self._get_partner(partner_id)
```
victoralmau: IMO there is no need for the `_get_partner()` method, you can do it directly there `self.env["res.partner"].browse(partner_id)`

hr_holidays_public/readme/CREDITS.md
```diff
@@ -1 +1 @@
-The migration of this module from 16.0 to 17.0 was financially supported by Camptocamp.
+The migration of this module from 17.0 to 18.0 was financially supported by Camptocamp.
```
victoralmau, line 2: This should not be changed.

hr_holidays_public/views/hr_holidays_public_view.xml
```diff
@@ -0,0 +1,64 @@
...
+                    <group name="group_main_right">
+                        <!-- Left empty for extensions -->
+                    </group>
+                </group>
+                <separator string="Public Holidays" name="group_detail" />
+                <field name="line_ids" nolabel="1">
+                    <list editable="top">
+                        <field
+                            name="date"
+                            readonly="variable_date == False"
```
victoralmau:
```suggestion
                            readonly="not variable_date"
```

victoralmau (changes requested):
Code and functional review ok.

Could yo squash administrative commits?

xaviedoanhduy:
hi @victoralmau, thanks for the suggestions

victoralmau (approved):

pedrobaeza:
We talk here about splitting the generic feature into a new module, and then provide an extension for HR:

https://github.com/OCA/hr-holidays/pull/110#issuecomment-2096096367

And what about the feature that Odoo has about the same concept?

cc @grindtildeath

victoralmau:
Please cherry-pick https://github.com/OCA/hr-holidays/pull/156 + https://github.com/OCA/hr-holidays/pull/157 to commit history.

My colleague's comment is still pending https://github.com/OCA/hr-holidays/pull/147#issuecomment-2459625077.

grindtildeath:
@pedrobaeza I wish I had more time to dive into this topic because the model exists since a few versions, but AFAICT standard public holidays still don't have the same flexibility as we have in the OCA module.

I guess a first step to converge both modules would be to have "rules" allowing creation of resource calendar leaves from the OCA holidays public.

Still, IMO would be nice to split the module to remove dependency on HR, and provide extension module for HR.

pedrobaeza:
Yes, we have talked today in my team about the Odoo core feature and it seems less complete than current one, and switching doesn't worth.

xaviedoanhduy:
hi i did some work:

- created PR for the extraction of the "public holiday" model to a new non-hr module, so that it could be used in other contexts: [[18.0][ADD] calendar_public_holiday](https://github.com/OCA/calendar/pull/142)
- prepared a new alternative PR for migration and refactoring of `hr_holidays_public` to depend on that new module:
[[18.0][MIG] hr_holidays_public: Migration to 18.0](https://github.com/OCA/hr-holidays/pull/159) 

let me know guys if this matches with your expectations

victoralmau:
Now that `calendar_public_holiday` already exists (https://github.com/OCA/calendar/pull/142), I believe that this PR should be modified to depend on `calendar_public_holiday` and only add the corresponding holidays. 
It would also be interesting migration scripts to convert the data: `calendar.event.type` and `hr.holidays.public` records to `calendar.public.holiday` (and lines).

xaviedoanhduy:
> Now that `calendar_public_holiday` already exists ([OCA/calendar#142](https://github.com/OCA/calendar/pull/142)), I believe that this PR should be modified to depend on `calendar_public_holiday` and only add the corresponding holidays. It would also be interesting migration scripts to convert the data: `calendar.event.type` and `hr.holidays.public` records to `calendar.public.holiday` (and lines).

hi @victoralmau, I've created a [alternative PR](https://github.com/OCA/hr-holidays/pull/159) to apply the new module dependency, but before I start working on the suggestion to write data migration scripts, I'm still not quite clear why `calendar.event.type` is there, or if you mean it should be `calendar.event` in the following code: https://github.com/OCA/hr-holidays/blob/e5d910b3eab4f83203602574e34ee2a91f0292a8/hr_holidays_public/models/hr_holidays_public.py#L146

victoralmau:
Can you modify this migration and add as a dependency `calendar_public_holiday` https://github.com/OCA/calendar/pull/142?

xaviedoanhduy:
> Can you modify this migration and add as a dependency `calendar_public_holiday` [OCA/calendar#142](https://github.com/OCA/calendar/pull/142)?

hi @victoralmau, i have created another PR to replace this PR in: https://github.com/OCA/hr-holidays/pull/159

victoralmau:
It seems that this PR can be closed in favor of https://github.com/OCA/hr-holidays/pull/159.

PR #149, opened 2024-11-14
[16.0][ADD] hr_holidays_natural_period_public

LauraCForgeFlow:
This glue module allows for the leave days computation in natural days to exclude public holidays.


victoralmau (changes requested):
IMO this is not necessary, because you can define whether to exclude public holidays or not with https://github.com/OCA/hr-holidays/blob/16.0/hr_holidays_public/models/hr_leave_type.py#L11.

LauraCForgeFlow:
I noticed that even when excluding public holidays with the _exclude_public_holidays_ field mentioned, the computation of the number of days in natural days did not properly exclude public holidays.

If we remove the changes in https://github.com/OCA/hr-holidays/pull/149/files#diff-ab3064ebe6a3018dc278f12ada4ba1e2e695dc08fef6a4ceb6f0c3c935b3ea24 and execute the tests, those that involve excluding public holidays fail (_test_hr_leave_natural_day_public_holiday_02_, _test_hr_leave_natural_day_public_holiday_weekend_02_). 

However, with the changes introduced in this PR, all tests pass successfully.

victoralmau:
Excuse me, one important thing. The `hr_holidays_natural_period` module is intended to use natural days, therefore, public holidays should NOT be excluded; if the user selects Monday through Friday, if Wednesday is a holiday it should NOT be excluded (IMO). What do you think @pedrobaeza?

pedrobaeza (changes requested):
As said by my colleague, the module was conceived to always compute natural days without excluding public holidays, as that's how all our companies that are using it work.

Anyway, the module can be extended to be compatible with such configuration. What is not correct is to change the current module dependency, nor adapting data through migration script. Due to both things, although we may do a "soft dependency" only injecting the context, that means the migration script and to trick the tests, I think the best to do in this case is that you create a glue module `hr_holidays_natural_period_public` that do the bridge, and install when you need natural periods, but excluding public holidays.

LauraCForgeFlow:
After taking a deeper look, I completely agree with you. I have changed my code into a new glue module as you requested. 
I hope everything is looking good now, and thanks for the explanation. :)

MiquelRForgeFlow (approved):

pedrobaeza (approved):
Thanks for considering the changes.

/ocabot merge nobump

PR #150, opened 2024-11-19
[15.0][IMP] hr_holidays_public: include public holidays in employee's _get_unusual_days

AaronHForgeFlow:
This is useful in the time off overview calendar, from:
image

to:
image

cc @forgeflow

### DIFF
--- a/hr_holidays_public/models/__init__.py
+++ b/hr_holidays_public/models/__init__.py
@@ -4,3 +4,4 @@ from . import hr_leave
 from . import hr_leave_type
 from . import hr_holidays_public
 from . import resource_calendar
+from . import hr_employee
--- a/hr_holidays_public/models/hr_employee.py
+++ b/hr_holidays_public/models/hr_employee.py
@@ -1,0 +1,17 @@
+from odoo import fields, models
+
+
+class HREmployee(models.Model):
+    _inherit = "hr.employee"
+
+    def _get_unusual_days(self, date_from, date_to=None):
+        res = super()._get_unusual_days(date_from, date_to=date_to)
+        if not self:
+            return res
+        domain = self.env["hr.leave"]._get_domain_from_get_unusual_days(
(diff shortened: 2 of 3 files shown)


JordiMForgeFlow (approved):
code review 👍🏼

MiquelRForgeFlow (approved):

pedrobaeza:
Not sure unusual days is for that. cc @victoralmau

victoralmau (changes requested):
I do not agree with this change. Although the `_get_unusual_days()` https://github.com/odoo/odoo/blob/3a28e5b0adbb36bdb1155a6854cdfbe4e7f9b187/addons/hr/models/hr_employee.py#L442 method perhaps does not clarify much, it is important to explain what is currently happening in the dashboard with the different options:
- Unusual days based on the employee's work schedule: they are shown in gray color.
ejemplo
- Time Off days: The overlay event is shown on the days that correspond.
ejemplo-ausencia
- Public holidays: The same event is shown as for time off days.

**IMO** I think it is clear that public holidays should be treated as holidays and not as non-working days.

victoralmau:
Extra comment: If we access from the “Time off”
 smart-button of an employee to the dashboard, the public holidays are shown in gray color.
ejemplo

AaronHForgeFlow:
> IMO I think it is clear that public holidays should be treated as holidays and not as non-working days.

Ok, my use case is that all public holidays are non-working days across the company, and in that case it made sense to show it as non working day for all,  but I understand that is not the most "generic" case.  I am closing this then and move the feature to another place. Thank you for the review.

victoralmau:
Sorry, I didn't agree with adding this as a general thing, maybe being something per company (disabled by default) with `res.config.settings` could be a good contribution. what do you all think?

pedrobaeza:
Even at company level, public holidays may vary depending on the employee locations, so I don't think this should be added this way, even with a configuration switch.

PR #153, opened 2024-12-04
[16.0][FIX][hr_holidays_natural_period] date min/max start/end

mileo:
If the user's configuration is in a timezone and the browser the number of days off is calculated incorrectly showing fractions.

image


pedrobaeza:
Solved in #167

PR #159, opened 2024-12-26
[18.0][MIG] hr_holidays_public: Migration to 18.0 

xaviedoanhduy:
### note

- depend applied on new module: https://github.com/OCA/calendar/pull/142
- most of the tests that were removed during this migration have been added to the `calendar_public_holiday` module.


victoralmau:
Now that `calendar_public_holiday` is already merged, can you delete the commit _[DON'T MERGE] Add test-requirements.txt_ and I'll check it?

hr_holidays_public/readme/USAGE.md
```diff
@@ -0,0 +1,24 @@
+For adding public holidays:
+
+1.  Go to the menu *Calendar \> Public Holidays \> Public Holidays*.
```
victoralmau:
```suggestion
1.  Go to the menu *Calendar \> Configuration > Public Holidays \> Public Holidays*.
```

victoralmau (changes requested):
Code and functional review OK.

This change is pending https://github.com/OCA/calendar/pull/142#issuecomment-2652989869

hr_holidays_public/readme/USAGE.md
xaviedoanhduy: hi @victoralmau, thanks for the suggestion, updated

victoralmau:
Comment https://github.com/OCA/calendar/pull/142#issuecomment-2652989869 is still pending.

xaviedoanhduy:
> Comment [OCA/calendar#142 (comment)](https://github.com/OCA/calendar/pull/142#issuecomment-2652989869) is still pending.

hi, I added a commit in response to that comment

victoralmau:
> > Comment [OCA/calendar#142 (comment)](https://github.com/OCA/calendar/pull/142#issuecomment-2652989869) is still pending.
> 
> hi, I added a commit in response to that comment

Can you add it in a .csv file to make it more readable? (similar to https://github.com/OCA/calendar/blob/18.0/calendar_public_holiday/security/ir.model.access.csv) I think this change does not need to be in an extra commit, it can be in the migration commit.

xaviedoanhduy:
> Can you add it in a .csv file to make it more readable? (similar to https://github.com/OCA/calendar/blob/18.0/calendar_public_holiday/security/ir.model.access.csv) I think this change does not need to be in an extra commit, it can be in the migration commit.

ok, updated

victoralmau (approved):
Code and functional review OK.

Ping @pedrobaeza

ArnauCForgeFlow (approved):

hr_holidays_public/models/hr_leave.py
```diff
@@ -23,37 +23,22 @@ def action_validate(self):
                 leave = leave.with_context(
                     employee_id=leave.employee_id.id, exclude_public_holidays=True
                 )
-            super(HrLeave, leave).action_validate()
+            super(HrLeave, leave).action_validate(check_state)
```
pedrobaeza: Always preserve keyword arguments:
```suggestion
            super(HrLeave, leave).action_validate(check_state=check_state)
```

pedrobaeza (commented):
I'm missing a menu entry for managing public holidays for HR managers.

xaviedoanhduy:
> I'm missing a menu entry for managing public holidays for HR managers.

hi, are you confused **between Time** off and **Employees**? because i brought it back to the original module where the time off manager has the right
image

hr_holidays_public/models/hr_leave.py
xaviedoanhduy: Thanks for your advice

pedrobaeza:
I'm talking about a menu entry. I see the security access rules, but not UI counterpart for managing the public holidays from inside the Time Off menus.

xaviedoanhduy:
added menu in manage public holidays inside the Time Off menus.

pedrobaeza:
Please put the menu inside Time-Off > Configuration, and I see there another "Public Holidays" menu from standard. Isn't they overlapping?

xaviedoanhduy:
> Please put the menu inside Time-Off > Configuration, and I see there another "Public Holidays" menu from standard. Isn't they overlapping?

yes, it is a standard feature of odoo "Public Holidays". i think that menu is outside all this time to avoid misunderstanding of duplication with that feature.
and actually those menus were in menu `Configuation`? I think I just put them back to their original place, see the code I have coppied from version `16.0`: https://github.com/OCA/hr-holidays/blob/bf3846d2f24adfc0269085a081d342733305f5c7/hr_holidays_public/views/hr_holidays_public_view.xml#L52-L63

CRogos (commented):
@xaviedoanhduy could you also cleanup the commit history. 

Squash administrative commits (if any) with the previous commit for reducing commit noise. Check https://github.com/OCA/maintainer-tools/wiki/Merge-commits-in-pull-requests#mergesquash-the-commits-generated-by-bots-or-weblate for details.

pedrobaeza:
@xaviedoanhduy please put it inside "Configuration" as talked. What we can do is to rename the menu to "Public Holidays (OCA)" to avoid duplication, but as it was it's not OK, because it creates another top level menu that is ugly, and it's also after "Configuration", which should be the last one.

xaviedoanhduy:
hello @pedrobaeza, Is this what you expected?

image

pedrobaeza (approved):
OK, that may serve.

/ocabot merge nobump

hr_holidays_public/models/hr_leave.py
```diff
@@ -23,37 +23,24 @@ def action_validate(self):
...
+        for leave in self:
+            if (
+                leave.holiday_status_id.exclude_public_holidays
+                or not leave.holiday_status_id
+            ):
+                leave = leave.with_context(
+                    employee_id=leave.employee_id.id, exclude_public_holidays=True
+                )
+            return super(HrLeave, leave)._get_durations(
+                check_leave_type=check_leave_type, resource_calendar=resource_calendar
```
marielejeune, line 39: Hi @xaviedoanhduy 

I know that this PR is merged but can you give me a bit of context of why you changed this please?

It causes errors because you are doing a return inside a for loop, hence returning the `_get_durations` result for only one record, not the whole recordset!

I would like to fix it asap (and I can do it) but if you have a bit of context about why you did this change, I would really appreciate.

xaviedoanhduy: > I would like to fix it asap (and I can do it) but if you have a bit of context about why you did this change, I would really appreciate.

hello @marielejeune, thank you for your feedback!

I'm curious to understand what kind of issue you encountered when `_get_durations()` was called with a multi-recordset — did it cause incorrect results or an exception?

In my case, I had tested this implementation and it returned the correct results in the existing test cases. I also tried passing a recordset to `super()` with a custom context (grouped by `exclude_public_holidays`), but that caused some test failures, possibly due to inconsistent behavior inside `_get_durations()` when handling multiple records with mixed leave types.

If you're able to reproduce the issue or propose a working fix that maintains test correctness, I’d be very happy to learn from it and improve the patch accordingly!

victoralmau: https://github.com/OCA/hr-holidays/pull/194

marielejeune: Hi @victoralmau , thanks a lot for the fix, I'm going to have a look.
@xaviedoanhduy `_get_durations` method is called in `_compute_duration` compute method on `hr.leave` model. 
It must return a dictionary whose keys are leave ids, and Odoo expects every leave of the recordset to have an entry in the dictionary. 
In this addon you extend the `_get_durations` method but in some cases you return after only 1 record, hence the dictionary is not fully completed. It can cause errors in many flows, I'm surprised you didn't encounter. 
More generally, an extension of a method should never change the format / specs of the output of the original method.

MiquelRForgeFlow:
Hi people. Have someone considered https://github.com/odoo/odoo/commit/0d846ecd1ec7ff5149d580d3494b1a4bab1e68d2 when migrating to v18? Any thoughts? Just asking.

PR #161, opened 2025-01-28
[16.0][ADD] hr_holidays_public_work_location

LauraCForgeFlow:
This module adds the work location of employees at the public vacation line level as an extra discriminant. That way, we can have some public vacations that are only valid for employees working in a specific location, even when there are more locations in the same country or state.

This PR depends on #164.


AaronHForgeFlow (approved):
Code LGTM!

BTW, It must be hard to not get a public holiday because you work in a specific work location :crying_cat_face:

MiquelRForgeFlow (approved):

AlexPForgeFlow (approved):

PR #164, opened 2025-02-10
[16.0][FIX] hr_holidays_public: keep 'employee_id' field amongst method calls

LauraCForgeFlow:
Before this commit, the `employee_id`'s parameter value was lost when calling some functions from other ones in this module (for example, when calling `_get_domain_states_filter` from `get_holidays_list`.

However, this parameter might be necessary in the computation of some holidays, when having restrictions related to the employee, instead of the partner (accessed through `employee.address_id`). You can see an example of that in the module added in #161.

### DIFF
--- a/hr_holidays_public/models/hr_holidays_public.py
+++ b/hr_holidays_public/models/hr_holidays_public.py
@@ -107,7 +107,11 @@ class HrHolidaysPublic(models.Model):
             return self.env["hr.holidays.public.line"]
         partner_id = partner.id if partner else None
         states_filter = self._get_domain_states_filter(
-            pholidays, start_dt, end_dt, partner_id=partner_id
+            pholidays,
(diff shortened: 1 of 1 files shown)


MiquelRForgeFlow (approved):

AaronHForgeFlow (approved):
code LGTM

AlexPForgeFlow (approved):
makes sense :+1:

GuillemCForgeFlow (approved):

AaronHForgeFlow:
ping @OCA/human-resources-maintainers

MiquelRForgeFlow:
Please, forward port.

LauraCForgeFlow:
@MiquelRForgeFlow on later versions this is already implemented, so no need to forward port. :)

PR #167, opened 2025-02-27
[16.0][FIX] hr_holidays_natural_period: Define the correct number of days in some use cases.

victoralmau:
Define the correct number of days in some use cases.

Example use case:
- Calendar Monday 16:00-21:00
- Calendar Tuesday-Friday: 09:00-14:00
- Absence from Monday to Sunday in natural day
- Number of days: 7

Please @pedrobaeza can you review it?

@Tecnativa TT55243

PR #168, opened 2025-02-28
[17.0][FIX] hr_holidays_natural_period: Define the correct number of days in some use cases

victoralmau:
FWP from 16.0: https://github.com/OCA/hr-holidays/pull/167

Define the correct number of days in some use cases.

Example use case:
- Calendar Monday 16:00-21:00
- Calendar Tuesday-Friday: 09:00-14:00
- Absence from Monday to Sunday in natural day
- Number of days: 7

This is due to the rrule we are using to generate the missing intervals for the non working days to simulate the natural period. Previously, we use the starting and ending time, resulting that the daily rrule was omitting the initial or ending day, as the theoretical hour is after the ending datetime. For the example above, the rrule will generate the entries from Monday at 16:00, so when the time for the Sunday rule comes, as Sun 16:00 is bigger than Sun 14:00, which was the ending datetime, no iteration is done for Sunday.

The solution is to put the rrule with the min time for the starting limit, and the maximum for the ending one.

It has been checked that there's no problem with timezones due to this patch.

@Tecnativa TT55243

### DIFF
--- a/hr_holidays_natural_period/models/resource_calendar.py
+++ b/hr_holidays_natural_period/models/resource_calendar.py
@@ -56,6 +56,10 @@ class ResourceCalendar(models.Model):
         return False
 
     def _natural_period_intervals_batch(self, start_dt, end_dt, intervals, resources):
+        # Re-define start_dt and end_dt to ensure that we always iterate through the
+        # last day.
+        start_dt = datetime.combine(start_dt.date(), time.min)
+        end_dt = datetime.combine(end_dt.date(), time.max)
         for resource in resources or []:
             interval_resource = intervals[resource.id]
             tz = timezone(resource.tz)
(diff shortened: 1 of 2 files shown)


pedrobaeza:
Something that is missing in the previous PR is the explanation of the problem, not only the steps to reproduce it. Please add to the commit message and PR description the following text after the steps:

```
This is due to the rrule we are using to generate the missing intervals for the non working days to simulate the natural period. Previously, we use the starting and ending time, resulting that the daily rrule was omitting the initial or ending day, as the theoretical hour is after the ending datetime. For the example above, the rrule will generate the entries from Monday at 16:00, so when the time for the Sunday rule comes, as Sun 16:00 is bigger than Sun 14:00, which was the ending datetime, no iteration is done for Sunday.

The solution is to put the rrule with the min time for the starting limit, and the maximum for the ending one.

It has been checked that there's no problem with timezones due to this patch.
```

victoralmau:
Ping @pedrobaeza

pedrobaeza (approved):
/ocabot merge patch

PR #169, opened 2025-03-17
[17.0][FIX] hr_holidays_natural_period: _get_duration handle modified request_unit

Tardo:
Solves a problem with the calculation of days. The problem comes from the “hack” that is used to calculate the days as “day” type (since the Odoo method is ungovernable). With this fix, a history is kept in the context to know if a record is really of type “day” or “natural_day”.

time_off

Steps to reproduce (with demo data):

1. Go to "Time Off > Configuration > Time Off Types" application
2. Edit "Paid Time Off" and change "Take Time Off in" to "Natural Day"
3. Go to calendar and allocate 1 week (7 natural days)
4. Check the "Paid Time Off" counter... it shows 15 instead of 13

### DIFF
--- a/hr_holidays_natural_period/README.rst
+++ b/hr_holidays_natural_period/README.rst
@@ -81,6 +81,10 @@ Contributors
 
      - Antoni Marroig <amarroig@apsl.net>
 
+- Grupo Isonor <https://www.grupoisonor.es>
+
+     - Alexandre D. Díaz
+
 Maintainers
 -----------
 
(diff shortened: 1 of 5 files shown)


torojur25 (approved):

pedrobaeza:
Please specify the steps to reproduce the problem, and the screenshot is not saying too much. Maybe it's the screenshot after the fix, but then, which was the screenshot before?

admingrupoisonores (approved):

Arubaru86 (approved):

Tardo:
@pedrobaeza Right, sorry for the lack of details. I have updated the first comment with the steps so that it can be easily reporduced.

victoralmau (approved):
Functional review OK

PR #172, opened 2025-04-07
[16.0][FIX]hr_holidays_public: handle incompatibility with `hr_work_entry_holidays` by propagating the context upwards

PicchiSeba:
The method `_get_number_of_days` introduced in `hr_work_entry_holidays`, is not being passed `exclude_public_holidays` in the context.
With this PR we propagate it in the calling methods too.


HekkiMelody (approved):
Code review, LGTM

HekkiMelody:
Hello @OCA/human-resources-maintainers , admittedly this is something of a hack but we've been using it in production for a couple of months now, and it works. Could you take a look?

PR #173, opened 2025-04-14
[17.0][IMP] hr_holidays_public: improve way to get public holidays

imlopes:
With this module we can create public holidays : 
* global, for any country
* by country
* by region/state

If we don't have an employee when calling `get_holidays_list`, we don't have the filter by states applied correctly.
This PR will just pass the ``resource.resource::employee_id`` in the context.

### DIFF
--- a/hr_holidays_public/models/resource_calendar.py
+++ b/hr_holidays_public/models/resource_calendar.py
@@ -45,6 +45,11 @@ class ResourceCalendar(models.Model):
             lunch=lunch,
         )
         if self.env.context.get("exclude_public_holidays") and resources:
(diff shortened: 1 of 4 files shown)


hr_holidays_public/models/resource_calendar.py
```diff
@@ -45,6 +45,8 @@ def _attendance_intervals_batch(
             lunch=lunch,
         )
         if self.env.context.get("exclude_public_holidays") and resources:
+            employee_id = resources.employee_id.id or False
+            self = self.with_context(employee_id=employee_id)
```
vvrossem: what happens in case of multiple records in `resources` ?

PR #142, opened 2024-09-17
[FIX] hr_holidays_public: calendar.get_work_hours_count

petrus-v:
Allow to call get_work_hours_count on resource.calendar taking care of public holidays

PR #174, opened 2025-04-15
[17.0][FIX] hr_holidays_public: calendar.get_work_hours_count

petrus-v:
Allow to call get_work_hours_count on resource.calendar taking care of public holidays.

This is a forward port of this PR https://github.com/OCA/hr-holidays/pull/142

This method returns the number of work hours between two datetimes.

It seems to be used by [odoo while computing theorical hours](https://github.com/OCA/OCB/blob/17.0/addons/hr_holidays/models/hr_leave.py#L588) in case of company hollidays (not related to an employee).

It's used by hr_attendance_validation (using this PR: https://github.com/OCA/hr-attendance/pull/180 on v14.0) used to compute the theoretical number of hours to works for the given date range.

I'm currently migrating those modules to v17.0

### DIFF
--- a/hr_holidays_public/models/resource_calendar.py
+++ b/hr_holidays_public/models/resource_calendar.py
@@ -2,6 +2,7 @@
 # Copyright 2018 Brainbean Apps
 # Copyright 2020 InitOS Gmbh
 # Copyright 2021 Tecnativa - Víctor Martínez
+# Copyright 2024 Pierre Verkest
 # License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
 
 from odoo import models
@@ -44,6 +45,8 @@ class ResourceCalendar(models.Model):
             tz=tz,
             lunch=lunch,
         )
+        if resources is None:
+            resources = [self.env["resource.resource"]]
         if self.env.context.get("exclude_public_holidays") and resources:
             return self._attendance_intervals_batch_exclude_public_holidays(
                 start_dt, end_dt, res, resources, tz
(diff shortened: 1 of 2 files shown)


hr_holidays_public/models/resource_calendar.py
```diff
@@ -44,6 +45,10 @@ def _attendance_intervals_batch(
             tz=tz,
             lunch=lunch,
         )
+
```
pedrobaeza: Please don't add empty lines inside a method.

pedrobaeza (commented):
Please indicate a use case for this done from user interface, as it's difficult to see the application of this change and sounds like a patch for custom code.

petrus-v:
Thanks for review @pedrobaeza I've remove the empty line and add description in the PR, hope it makes it makes things clear ?

hr_holidays_public/models/resource_calendar.py
```diff
@@ -44,6 +45,9 @@ def _attendance_intervals_batch(
             tz=tz,
             lunch=lunch,
         )
+        if resources is None:
+            resources = [self.env["resource.resource"]]
+
```
pedrobaeza: Still an empty line here.

pedrobaeza (commented):
Thanks for the extra info, but I'm still struggling why this module should do that "variable initialization" and in which case this provokes a problem.

hr_holidays_public/models/resource_calendar.py
petrus-v: fixed, my bad sorry about that :face_in_clouds: and again thanks for your time

PR #191, opened 2025-05-20
[16.0][FIX] hr_holidays_natural_period: Hide Half day option in leaves if natural_day is selected

victoralmau:
Hide _Half day_ option in leaves if natural_day is selected

**Before**
antes

**After**
despues

Please @pilarvargas-tecnativa and @pedrobaeza can you review it?

@Tecnativa TT56402

### DIFF
--- a/hr_holidays_natural_period/__manifest__.py
+++ b/hr_holidays_natural_period/__manifest__.py
@@ -13,4 +13,5 @@
     "depends": ["hr_holidays"],
     "maintainers": ["victoralmau"],
     "demo": ["demo/hr_leave_type_data.xml"],
+    "data": ["views/hr_leave_views.xml"],
 }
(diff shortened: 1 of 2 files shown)


pedrobaeza:
It's correct to request half day using natural days.

victoralmau:
I am not so sure that this is correct, if you try to create a leave for a type that has defined days it is not possible to define half days, but with the option of “natural days” yes, it is strange to me, but I guess it is clear that it is correct as it is.

pedrobaeza:
Well, both are valid. You may add two options: natural days with half days, and only natural days following the same pattern than the core one.

victoralmau:
> Well, both are valid. You may add two options: natural days with half days, and only natural days following the same pattern than the core one.

Ok, can you reopen this to apply those changes here?

pedrobaeza:
Oh, sorry, GitHub is not allowing me that. Open a new one.

PR #196, opened 2025-06-23
[16.0][IMP] hr_holidays_public: Show leave icon (airplane) if employee/user has a public holiday

victoralmau:
Show leave icon (airplane) if employee/user has a public holiday

ejemplo-1
ejemplo-2

Please @pedrobaeza and @carlos-lopez-tecnativa can you review it?

@Tecnativa TT56844

### DIFF
--- a/hr_holidays_public/models/hr_employee.py
+++ b/hr_holidays_public/models/hr_employee.py
@@ -1,0 +1,33 @@
+# Copyright 2025 Tecnativa - Víctor Martínez
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
+
+from odoo import fields, models
+
+
+class HrEmployeeBase(models.AbstractModel):
+    _inherit = "hr.employee.base"
+
+    is_public_holiday = fields.Boolean(
+        string="Public Holiday Today", compute="_compute_is_public_holiday"
+    )
+
+    def _compute_is_public_holiday(self):
+        holiday_public = self.env["hr.holidays.public"]
+        for item in self:
+            item.is_public_holiday = holiday_public.is_public_holiday(
+                fields.Date.context_today(item), employee_id=item.id
+            )
+
+    def _get_im_status_hr_holidays_public(self, key):
+        im_status_mapped = {
+            "online": "leave_online",
+            "away": "leave_away",
+            "offline": "leave_offline",
+        }
+        return im_status_mapped[key]
+
+    def _compute_leave_status(self):
+        res = super()._compute_leave_status()
+        for item in self.filtered(lambda x: not x.is_absent and x.is_public_holiday):
+            item.is_absent = True
+        return res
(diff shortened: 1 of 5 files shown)


hr_holidays_public/models/res_users.py
```diff
@@ -0,0 +1,38 @@
...
+    _inherit = "res.users"
+
+    is_public_holiday = fields.Boolean(
+        string="Public Holiday Today", compute="_compute_is_public_holiday"
+    )
+
+    def _compute_is_public_holiday(self):
+        holiday_public = self.env["hr.holidays.public"]
+        for item in self:
+            employees = item.employee_ids
```
pedrobaeza: This should be filtered by the current env company, not the first one.

pedrobaeza (commented):
Can't you concentrate the logic a bit in the employee for not repeating code?

hr_holidays_public/models/res_users.py
victoralmau: I think doing that will have side effects, but OK.

pedrobaeza: The same of this, as the public holiday may not apply to the first employee. The other alternative is to check the public holidays for all the employees.

hr_holidays_public/models/hr_employee.py
```diff
@@ -0,0 +1,33 @@
...
+
+
+class HrEmployeeBase(models.AbstractModel):
+    _inherit = "hr.employee.base"
+
+    is_public_holiday = fields.Boolean(
+        string="Public Holiday Today", compute="_compute_is_public_holiday"
+    )
+
+    def _compute_is_public_holiday(self):
```
pedrobaeza, line 14: I think this will have cache persistence problems if you don't restart the server/worker. Is this needed to be a computed field? You can simply call a method in the `filtered` instead.

victoralmau: You mean remove that field and use an `_is_public_holiday()` method in `hr.employee` that would be used everywhere (`hr.employee`, `res.users` and `res.partner`)? I used the technique of creating a field because it is something that `hr_holidays` does with `is_absent` field https://github.com/odoo/odoo/blob/9728a1754424b9e5bfa93ec5d53a5179fa9d7d05/addons/hr_holidays/models/hr_employee.py#L41C89-L41C112

pedrobaeza: Well, I'm not sure about its persistence, and they have also implemented a `search` method that minimizes the problem. Let's try this way.

pedrobaeza (approved):

hr_holidays_public/models/hr_employee.py
```diff
@@ -0,0 +1,33 @@
...
+        holiday_public = self.env["hr.holidays.public"]
+        for item in self:
+            item.is_public_holiday = holiday_public.is_public_holiday(
+                fields.Date.today(), employee_id=item.id
```
carlos-lopez-tecnativa: Please use `context_today` to avoid issues with negative UTC offsets.
```suggestion
                fields.Date.context_today(item), employee_id=item.id
```

carlos-lopez-tecnativa (changes requested):
LGTM, just a minor change please

carlos-lopez-tecnativa (approved):

PR #203, opened 2025-07-17
[18.0][MIG] hr_holidays_validator_list: Migration to 18.0

Kimkhoi3010:
Migrate `hr_holidays_validator_list` to 18.0
### Change:
- `activity_update()` and `_check_approval_update()` methods internally use the `_get_responsible_for_approval()` method to determine the approvers. Therefore, instead of overriding these methods, we only need to extend `_get_responsible_for_approval()` to return the list of approvers from the `leave_manager_ids` field. This allows the system to automatically assign the correct approvers without further customization.


hr_holidays_validator_list/__manifest__.py
```diff
@@ -0,0 +1,19 @@
+# Copyright 2023 Camptocamp SA
```
cyrilmanuel, line 1: 2025

cyrilmanuel (approved):
lgtm

Ricardoalso (approved):
technical migration LGTM

hr_holidays_validator_list/models/hr_employee.py
```diff
@@ -0,0 +1,48 @@
...
+
+    def _add_leave_manager_ids_in_group(self):
+        approver_group = self.env.ref(
+            "hr_holidays.group_hr_holidays_responsible", raise_if_not_found=False
+        )
+        if not approver_group:
+            return
+        managers = self.mapped("leave_manager_ids")
+        if managers:
+            approver_group.users |= managers
```
victorvermot:
```suggestion
            approver_group.sudo().users |= managers
```
@Kimkhoi3010  You will need to use sudo here otherwise you get an access error on res.groups if you're not an admin user. Could you fix the pr ? Thanks

Kimkhoi3010: Hi @victorvermot, I have updated the PR.
Thanks for your time.

hr_holidays_validator_list/tests/test_hr_leave.py
```diff
@@ -0,0 +1,71 @@
...
+class TestHrLeave(TransactionCase):
+    def setUp(self):
+        super().setUp()
+        employee = self.env["hr.employee"].search([])[0]
+        employee.write({"leave_manager_ids": [(6, 0, [6, 2])]})
+        hr_leave_type = self.env["hr.leave.type"].create(
+            {
+                "name": "Leave Test",
+                "request_unit": "hour",
+            }
```
victorvermot, line 20:
```suggestion
            "requires_allocation": "no",
```

Your tests are failing because you create an hr leave with a type that requires to create an allocation beforehand. Add this line to say that it does not require it

Kimkhoi3010: Hi @victorvermot,  
I have updated the PR. However, the tests are failing in the module `hr_holidays_natural_period`.
Thanks for your time.

cyrilmanuel:
@Kimkhoi3010  can you retrigger the build and let me know if you need some help ?

PR #211, opened 2025-08-13
[17.0][IMP] hr_holidays_natural_period: reimplementation

ofonseca-pyming:
When contracts changed to 'open' state with natural_day leave types, the system
  would hang for minutes due to cascading field recomputations. The root cause was
  the original implementation's approach of temporarily modifying the
  holiday_status_id.request_unit field from 'natural_day' to 'day' during duration
  calculations.

  The issue is that this system hangs for minutes when changing contract states 
  with natural_day leave types.  Install hr_contract and try to validate a contract after 
  setting leave time "Paid Time Off" to natural days.

  - Complete reimplementation of _get_duration without field modifications
  - Take into account resource.calendar.leaves and hr.holidays.public
  - Eliminates performance issue architecturally rather than with caching workarounds
  - Added more tests
 
Grabación de pantalla desde 2025-08-19 11-32-42-optimized


hr_holidays_natural_period/tests/common.py
```diff
@@ -0,0 +1,151 @@
...
+    """Mixin providing common test utilities for natural period tests."""
+
+    @classmethod
+    def setUpClass(cls):
+        """Set up minimal shared test data."""
+        super().setUpClass()
+        cls.env = cls.env(context=dict(cls.env.context, **DISABLED_MAIL_CONTEXT))
+        cls.company = cls.env.ref("base.main_company")
+        cls.calendar = cls.env.ref("resource.resource_calendar_std")
+        cls.env.user.tz = "UTC"
```
victoralmau, line 23:
```suggestion
from odoo.addons.base.tests.common import BaseCommon


class HrLeaveNaturalPeriodTestMixin(BaseCommon):
    """Mixin providing common test utilities for natural period tests."""

    @classmethod
    def setUpClass(cls):
        """Set up minimal shared test data."""
        super().setUpClass()
        cls.env = cls.env(context=dict(cls.env.context, **DISABLED_MAIL_CONTEXT))
        cls.company = cls.env.ref("base.main_company")
        cls.calendar = cls.env.ref("resource.resource_calendar_std")
        cls.natural_leave_type = cls._create_natural_day_leave_type()
        cls.regular_leave_type = cls._create_regular_day_leave_type()
        cls.employee = cls._create_test_employee()
        cls.env.user.tz = "UTC"
```

hr_holidays_natural_period/tests/test_hr_leave.py
```diff
@@ -1,24 +1,30 @@
...
+from odoo import Command, fields
 from odoo.tests import Form, new_test_user
 from odoo.tests.common import users
+from odoo.tools import mute_logger
 
 from odoo.addons.base.tests.common import BaseCommon
 
 
-@freeze_time("2023-01-01", tick=True)
 class TestHrLeave(BaseCommon):
```
victoralmau, line 14: Perhaps `HrLeaveNaturalPeriodTestMixin` could be extended here, although I'm not sure if it adds much.

hr_holidays_natural_period/models/__init__.py
```diff
@@ -1,5 +1,6 @@
...
-from . import hr_leave
-from . import resource_calendar
-from . import hr_leave_allocation
-from . import hr_employee
+from . import (
+    hr_leave_type,
+    hr_leave,
+    hr_leave_allocation,
+    hr_employee,
+)
```
victoralmau, line 6: IMO, there is no need to change this, it generates an unnecessary diff.

hr_holidays_public/__manifest__.py
```diff
@@ -4,7 +4,7 @@
 
 {
     "name": "HR Holidays Public",
-    "version": "17.0.1.0.7",
+    "version": "17.0.1.0.8",
```
victoralmau, line 7:
```suggestion
```
This change is not necessary; the bot will do it automatically.

The _[BOT] post-merge updates_ commit is not really necessary.

hr_holidays_natural_period/models/hr_leave.py
```diff
@@ -3,26 +3,230 @@
...
+from odoo import fields, models
+from odoo.tools.date_utils import end_of, start_of
+from odoo.tools.float_utils import float_round
+
+from odoo.addons.resource.models.utils import HOURS_PER_DAY
 
 
 class HrLeave(models.Model):
+    """Extensions to handle natural day leave durations."""
+
```
victoralmau, line 19:
```suggestion
```
This should not be done because it overwrites the “_base_” docstring. https://github.com/odoo/odoo/blob/cc0060e889603eb2e47fa44a8a22a70d7d784185/addons/hr_holidays/models/hr_leave.py#L31

hr_holidays_natural_period/models/hr_leave_allocation.py
```diff
@@ -1,10 +1,14 @@
 # Copyright 2024 APSL-Nagarro - Antoni Marroig Campomar
 # License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
 
+"""Natural period: extend allocation with 'Natural day' request unit."""
+
```
victoralmau, line 5:
```suggestion
```

hr_holidays_natural_period/models/hr_leave_allocation.py
```diff
@@ -1,10 +1,14 @@
...
 from odoo import fields, models
 
 
 class HrLeaveAllocation(models.Model):
+    """Extend allocation to support natural day request unit."""
+
```
victoralmau, line 11:
```suggestion
```
Similar to what was mentioned above.

hr_holidays_natural_period/models/hr_leave_type.py
```diff
@@ -1,10 +1,14 @@
 # Copyright 2020-2023 Tecnativa - Víctor Martínez
 # License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
 
+"""Natural period: extend leave type with 'natural_day' request unit."""
+
```
victoralmau, line 5:
```suggestion
```

hr_holidays_natural_period/models/hr_leave_type.py
```diff
@@ -1,10 +1,14 @@
...
 from odoo import fields, models
 
 
 class HrLeaveType(models.Model):
+    """Extend leave type with 'natural_day' request unit."""
+
```
victoralmau, line 11:
```suggestion
```
Similar to what was mentioned above.

hr_holidays_natural_period/tests/__init__.py
```diff
@@ -1 +1,8 @@
-from . import test_hr_leave
+from . import (
+    test_hr_leave,
+    test_hr_leave_natural_day_detection,
+    test_hr_leave_duration_calculation,
+    test_hr_leave_integration,
+    test_hr_leave_performance,
+    test_public_holiday_exclusion,
```
victoralmau, line 7: Are all these tests necessary to test the use case you describe? Perhaps too many things are being tested (although I haven't reviewed all the tests in detail).

hr_holidays_natural_period/models/hr_employee.py
```diff
@@ -9,16 +9,23 @@ class HrEmployee(models.Model):
...
+        if natural_day_types:
+            # Use context to signal natural day calculation without field modification
+            ctx = dict(
+                self.env.context,
+                mod_holidays_status_ids=natural_day_types.ids,
+                natural_day_computation=True,
+            )
+            return super(HrEmployee, self.with_context(**ctx))._get_consumed_leaves(
+                leave_types, target_date, ignore_future
+            )
```
victoralmau, line 30:
```suggestion
        self = self.with_context(mod_holidays_status_ids=natural_day_types.ids)
```
I think this would be more readable.

victoralmau (changes requested):
Are all these changes really necessary? I find it very difficult to understand the changes and, above all, why they are necessary.
Couldn't the “_problem_” be solved in a simpler way than rewriting a lot of logic that `hr_holidays` already has? (For example, using some context so that when the contract status changes, it doesn't take so long to recalculate.)

PR #216, opened 2025-09-25
[16.0][FIX] hr_holidays_public: Apply the appropriate domain to display public holidays

victoralmau:
Apply the appropriate domain to display public holidays

Example use case:
- Define two dates as public holidays: 2025-12-30 + 2025-12-31
- Go to the Time off dashboard
- Both dates (2025-12-30 + 2025-12-31) should be displayed as Public Holidays

Please @pedrobaeza and @david-banon-tecnativa can you review it?

@Tecnativa TT58183

### DIFF
--- a/hr_holidays_public/models/hr_leave.py
+++ b/hr_holidays_public/models/hr_leave.py
@@ -63,13 +63,7 @@ class HrLeave(models.Model):
             else self.env.user.employee_id
         )
         if date_to:
-            domain.append(
-                (
-                    "date",
-                    "<",
-                    date_to,
-                )
-            )
+            domain.append(("date", "<=", date_to))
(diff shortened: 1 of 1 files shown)


hr_holidays_public/models/hr_leave.py
```diff
@@ -66,7 +66,7 @@ def _get_domain_from_get_unusual_days(self, date_from, date_to=None):
             domain.append(
```
pedrobaeza:
```suggestion
            domain.append(("date", "<=", date_to))
```

pedrobaeza (commented):
Let's reduce useless lines

pedrobaeza (approved):

david-banon-tecnativa (approved):

PR #225, opened 2025-10-21
[17.0][FIX] hr_holidays_public: calendar.get_work_hours_count

petrus-v:
re-open #174 

Allow to call get_work_hours_count on resource.calendar taking care of public holidays

This is a forward port of this PR https://github.com/OCA/hr-holidays/pull/142

This method returns the number of work hours between two datetimes.

It seems to be used by [odoo while computing theorical hours](https://github.com/OCA/OCB/blob/17.0/addons/hr_holidays/models/hr_leave.py#L588) in case of company hollidays (not related to an employee).

It's used by hr_attendance_validation (using this PR: https://github.com/OCA/hr-attendance/pull/180 on v14.0) used to compute the theoretical number of hours to works for the given date range.

I'm currently migrating those modules to v17.0

### DIFF
--- a/hr_holidays_public/models/resource_calendar.py
+++ b/hr_holidays_public/models/resource_calendar.py
@@ -2,6 +2,7 @@
 # Copyright 2018 Brainbean Apps
 # Copyright 2020 InitOS Gmbh
 # Copyright 2021 Tecnativa - Víctor Martínez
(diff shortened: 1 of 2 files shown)


pedrobaeza (approved):
/ocabot merge patch

PR #226, opened 2025-11-03
[18.0][FIX] hr_holidays_natural_period: not expect a singleton

Tardo:
"self" can be a recordset with more than one record. This can generate a recordset with more than one record for "self.holiday_status_id" This causes a singleton exception when retrieving the “request_unit” value. This also causes problems when sending the "original/old" value to the context... because it will be applying the same context to all "holiday_status_id" that may have different ‘request_unit’.

### DIFF
--- a/hr_holidays_natural_period/models/hr_leave.py
+++ b/hr_holidays_natural_period/models/hr_leave.py
@@ -2,6 +2,8 @@
 # Copyright 2024 Tecnativa - Carlos Lopez
 # Copyright 2025 Grupo Isonor - Alexandre D. Díaz
 # License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
+from collections import defaultdict
+
 from odoo import models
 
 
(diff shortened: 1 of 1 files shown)


victoralmau (approved):
It seems like a logical change, thanks.

hr_holidays_natural_period/models/hr_leave.py
```diff
@@ -12,20 +14,39 @@ def _get_durations(self, check_leave_type=True, resource_calendar=None):
...
             in ("natural_day", "natural_day_half_day")
         )
-        natural_day_instances.holiday_status_id.sudo().request_unit = (
-            "half_day" if old_request_unit == "natural_day_half_day" else "day"
-        )
+        orig_status_request_units = {
+            leave.holiday_status_id.id: leave.holiday_status_id.request_unit
+            for leave in self
+        }
+
```
pedrobaeza: Do this only for the filtered recordset, and don't add empty lines inside the method.

pedrobaeza (changes requested):

cyrilmanuel:
hi @Tardo nice catch :)

etobella:
@Tardo What is missing on this?

Tardo:
Updated with the suggested changes and made a small refactoring so that it only works with natural_day_instances

Tardo:
Ok, ready :D

Tardo:
I'm sorry for all the changes... but I was only taking into account the "request_unit" without the holiday_status_id... now fixed.

victoralmau:
Ping @pedrobaeza Can you review it?

pedrobaeza (approved):
/ocabot merge patch

PR #228, opened 2025-11-05
[IMP] create module resource_calendar_flexible_exclude_weekend on odoo 18.0

cyrilmanuel:
In the schedule  “flexible hours”, the work calendar has a number of hours per week, and this total number of hours is set manually (as is the average number of hours per day. What we expect is that when an employee with such a work schedule creates a leave request, the week ends are not counted (which is not the case, for Odoo, flexible hours does not exclude week ends).


resource_calendar_flexible_hours_exclude_weekend/models/resource_calendar.py
```diff
@@ -0,0 +1,142 @@
...
+            for resource in resources_with_flex_no_weekend:
+                resources_per_tz[tz or timezone((resource or self).tz)].append(resource)
+            for tz, tz_resources in resources_per_tz.items():
+                skipping_start_dt = start_dt.astimezone(tz)
+                # 0=Monday, 6=Sunday
+                weekday = skipping_start_dt.weekday()
+                if weekday in (5, 6):  # Saturday or Sunday
+                    # Move to next Monday, midnight in the considered TZ
+                    days_to_monday = 7 - weekday
+                    skipping_start_dt += timedelta(days=days_to_monday)
```
gurneyalex:
```suggestion
                    skipping_start_dt += timedelta(days=days_to_monday, hours=1)
```

Handle the fact that there could be a change from non DST to DST period in the interval

resource_calendar_flexible_hours_exclude_weekend/models/resource_calendar.py
```diff
@@ -0,0 +1,142 @@
...
+                    )
+
+                # for resources which should skip weekend we have to iterate by week
+                # XXX Maybe we can be smarter and only do 1 or 2 calls:
+                # 1 call if the first day is a monday, and a second if we need to
+                # check the following week.
+                while skipping_start_dt < end_dt:
+                    # find the end of the current week or the end of the period
+                    skipping_end_dt = skipping_start_dt + timedelta(
+                        days=7 - skipping_start_dt.weekday()
```
gurneyalex:
```suggestion
                        days=7 - skipping_start_dt.weekday(),
                        hours=1,
```

Same as above

gurneyalex (commented):
Can you try with this?

resource_calendar_flexible_exclude_weekend/models/resource_calendar.py
```diff
@@ -0,0 +1,142 @@
...
+        - If Saturday or Sunday → move to next Monday
+        - Then call super() with the adjusted start_dt
+        """
+        # Ensure timezone awareness
+        assert start_dt.tzinfo and end_dt.tzinfo, "Datetimes must be timezone-aware"
+        self.ensure_one()
+
+        if not resources:
+            resources = self.env["resource.resource"]
+            resources_list = [resources]
```
Ricardoalso, line 38:
```suggestion
            resources_list = resources
```

Why transforming this empty recordset into a list ?

Ricardoalso: After looking into it further, it appears to originate from the original method defined in the resource addon. It might be better to simplify the logic rather than copying the complex code from there.

Ricardoalso (commented):
By curiosity, was the configuration for Mandatory Days explored in order to automatically exclude leave calculations ? It demands the creation of  _hr.leave.mandatory.day_ records for each weekend, but it could do the work 🤔 

One interesting functional resource I found https://tenthplanet.in/odoo/product/time-off/odoo-18-time-off/

cyrilmanuel (commented):
LGTM, tested and work  !

wouitmil (approved):

wouitmil (approved):

santostelmo (approved):

cyrilmanuel:
hi @pedrobaeza  could you merge this pr please ?

resource_calendar_flexible_hours_exclude_weekend/readme/DESCRIPTION.md
```diff
@@ -0,0 +1 @@
+This module allow to compute the resource calendar as flexible =true, without taking the week-end days on the computation.
```
pedrobaeza:
```suggestion
This module allows to compute the resource calendar as flexible, without taking the week-end days on the computation.
```

pedrobaeza (commented):
The module name should avoid plurals, and also very long names, so what do you think about calling it `resource_calendar_flexible_exclude_weekend`?

resource_calendar_flexible_hours_exclude_weekend/readme/DESCRIPTION.md
cyrilmanuel: i agree :) thanks for your proposition ! 

it's done

pedrobaeza:
Thanks, can you squash the commits?

cyrilmanuel:
sure :)
it's done

PR #235, opened 2025-12-22
[18.0][FIX] hr_holidays_leave_repeated: handle DST timezones

remi-filament:
Before this fix, when using a timezone with DST and trying to repeat holidays over a period where DST changed (for instance if you want to repeat a holiday every friday over the full year), you would only get holidays before DST changed and after it was set back to original value.

This PR fixes that behaviour and adds corresponding tests (which were failing before the fix and solved afterwards)

### DIFF
--- a/hr_holidays_leave_repeated/models/hr_leave.py
+++ b/hr_holidays_leave_repeated/models/hr_leave.py
@@ -29,15 +29,30 @@ class HrLeave(models.Model):
 
     @api.model
     def _update_repeated_workday_dates(self, resource_calendar, from_dt, to_dt, days):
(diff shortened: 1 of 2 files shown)


remi-filament:
Hi @hailangvn do you mind having a look at this PR ?

hailangvn (approved):
LGTM. Thank you.

julienCozo (approved):

Hugo-Trentesaux (approved):

hildickethan (approved):

BhaveshHeliconia (approved):

PR #188, opened 2025-05-06
[ADD] hr_leave_date_to_custom_hour

SirPyTech:
Specify an end date when creating a leave with custom hours.

PR #236, opened 2025-12-23
[18.0][MIG] hr_leave_date_to_custom_hour: migration to 18.0

quirino95:
Migration of https://github.com/OCA/hr-holidays/pull/188 to 18.0


CRogos:
@quirino95  you need to preserve the commit history. 
See the migration guide: https://github.com/OCA/maintainer-tools/wiki/Migration-to-version-19.0

SirPyTech (changes requested):
**chore**:
> @quirino95 you need to preserve the commit history.
> See the migration guide: https://github.com/OCA/maintainer-tools/wiki/Migration-to-version-19.0

(from https://github.com/OCA/hr-holidays/pull/236#issuecomment-3686890146)

khanhratdeptrai (approved):

quirino95:
Hi @CRogos @SirPyTech,
Commit history has been updated, thank you!

hr_leave_date_to_custom_hour/models/hr_leave.py
```diff
@@ -15,11 +15,15 @@ class HRLeave(models.Model):
...
     def _compute_date_from_to(self):
-        return super()._compute_date_from_to()
-
-    def _get_attendances(self, employee, request_date_from, request_date_to):
-        if len(self) == 1 and self.request_unit_hours:
-            self.request_date_to = request_date_to = (
-                self.custom_hours_request_date_to or self.request_date_to
+        res = super()._compute_date_from_to()
+        if (
+            len(self) == 1
```
SirPyTech: **issue**: This `len(self) == 1` check was needed in `16.0` because the method `_get_attendances` was passed values specific to one record (`employee`, `request_date_*`) so it only made sense to change the behavior when `len(self) == 1`; note that it was also overwriting the parameter `request_date_to`.

It was still working correctly for all the records in the loop executed by `_compute_date_from_to` because `_get_attendances` was called on each record of the loop; after this PR, this would only work if such loop is one record long.

In order to fix it, since we are directly in the loop, we can remove this check and do another loop to overwrite the values computed in `super`; something like:
```py
for holiday in self:
    if (...):  #  same condition but using `holiday` and without len(self) == 1
        holiday.date_to = ...  # same value but using `holiday`
```
What do you think?

hr_leave_date_to_custom_hour/tests/test_hr_leave.py
SirPyTech: **thought**: The module `hr_leave_custom_hour_interval` does not exist in `18.0` so we could remove the `if` branches that are only there to support it, what do you think?

hr_leave_date_to_custom_hour/views/hr_leave_views.xml
```diff
@@ -9,25 +9,21 @@
...
+            >
+                <field name="custom_hours_request_date_to" invisible="1" />
+            </xpath>
+            <xpath
+                expr="//div[hasclass('o_row')]/field[@name='request_date_from']"
+                position="attributes"
+            >
                 <attribute name="widget">daterange</attribute>
                 <attribute name="options">{
-                    'related_end_date': 'custom_hours_request_date_to',
```
SirPyTech: **polish**: The trailing comma is there so that when adding/removing lines, the existing lines are not edited: it helps to keep the changes history clean.
But it's probably just my opinion so it's fine to remove it too.

SirPyTech (changes requested):

hr_leave_date_to_custom_hour/models/hr_leave.py
quirino95: Yes, that's a good fix, updated code.

hr_leave_date_to_custom_hour/tests/test_hr_leave.py
quirino95: Yes, now that method is much more readable

hr_leave_date_to_custom_hour/views/hr_leave_views.xml
quirino95: Is okay for me to keep the comma: code updated.

hr_leave_date_to_custom_hour/models/hr_leave.py
SirPyTech: :+1:

hr_leave_date_to_custom_hour/tests/test_hr_leave.py
SirPyTech: :+1:

hr_leave_date_to_custom_hour/views/hr_leave_views.xml
SirPyTech: :+1:

SirPyTech (approved):
Code and functional review, looks good 👍

anajuaristi:
Hi @CRogos @SirPyTech, any update on this? 
Could it be merged, please?
Thank you!

SirPyTech:
> Hi @CRogos @SirPyTech, any update on this? Could it be merged, please? Thank you!

I don't have merge permission, please mention the appropriate OCA team: you can find it in https://oca.github.io/repo-maintainer-conf/.

BhaveshHeliconia:
@quirino95 Could you please rebase the PR to make Runbot available?

BhaveshHeliconia (approved):
LGTM!

PR #237, opened 2026-01-05
[18.0][FIX] hr_holidays_public: Avoid failing computation of im_status

grindtildeath:
In case module hr_homeworking is installed, we can have different values for field res.users.im_status than those defined in function hr.employee.base._get_im_status_hr_holidays_public.

Therefore, any computation with such a value (eg presence_home_online) would result in a KeyError since the dict keys are accessed directly.

By using get and providing the same value as default in case the actual im_status is not defined we ensure the computation will not fail, even if any other module defines potentially new im_status values.


pedrobaeza (approved):

CRogos (approved):
LGTM

PR #238, opened 2026-01-13
[18.0][FIX] hr_holidays_natural_period: Performance problem on number_of_days

pedrobaeza:
When opening an employee that has leaves with natural days, and having a lot of leaves of this type for all the employees, there's a serious performance problem, because changing the request unit recomputes all the leaves of that type.

Using the trick from #60256 of updating just the cache with the desired value for the computations, avoid heavy recomputations.
    
This has uncovered also a problem of when the request unit is injected, so we reduce it to the minimum to avoid cross-changes.
    
@Tecnativa TT60256


victoralmau (approved):

pedrobaeza:
This is not yet working for new leaves. Working on a solution.

pedrobaeza:
OK, now is working restricting to the leave types not already touched.

Merging this for avoiding the problem:

/ocabot merge patch

PR #241, opened 2026-02-25
[18.0][FIX] hr_holidays_public_city: Add domain city_ids (similar to v17)

victoralmau:
Add domain `city_ids` (similar to v17)

Forgotten in https://github.com/OCA/hr-holidays/commit/37538ecb0fa618562fed5110dbace0f1ff47c28f

@Tecnativa TT61153


pedrobaeza (commented):
Can you add a regression test for not having this problem again?

pedrobaeza (approved):
Thanks for the test. Fast-tracking it for fixing the bug.

/ocabot merge patch

PR #243, opened 2026-03-25
[17.0][ADD]hr_holidays_approval_split

GuillermoNforgeflow:


hr_holidays_approval_split/views/hr_leave_type_views.xml
```diff
@@ -0,0 +1,16 @@
...
+    <record id="edit_holiday_status_form" model="ir.ui.view">
+        <field name="name">hr.leave.type.form.new.approval.field</field>
+        <field name="model">hr.leave.type</field>
+        <field name="inherit_id" ref="hr_holidays.edit_holiday_status_form" />
+        <field name="arch" type="xml">
+            <xpath expr="//field[@name='leave_validation_type']" position="after">
+                <field
+                    name="leave_validation_split"
+                    invisible="leave_validation_type != 'hr'"
+                />
```
GuillemCForgeFlow, line 12: why are the other approval options skipped? why only `hr`?

Image

hr_holidays_approval_split/models/hr_leave_type.py
```diff
@@ -0,0 +1,18 @@
...
+            ("both", "Hours and Days require approval"),
+            ("hours", "Only Hours require approval"),
+            ("days", "Only Days require approval"),
+        ],
+        string="Approval Type",
+        help="""
+            Define whether approval is required for hour-based requests,
+            day-based requests or both.
+        """,
+    )
```
GuillemCForgeFlow, line 19: I think we should have a default as both

hr_holidays_approvals/models/hr_leave.py
```diff
@@ -0,0 +1,41 @@
+from odoo import models
+
+
+class HolidaysRequest(models.Model):
+    _inherit = "hr.leave"
+
+    def activity_update(self):
```
GuillemCForgeFlow: `activity_update` doesn't seems to be the place to inherit, this method is called within other `action...` methods such as `action_validate`. this is also dangerous as it can create recursive calls.
perhaps we should use the `action_confirm`

hr_holidays_approvals/models/hr_leave.py
```diff
@@ -0,0 +1,41 @@
...
+                )
+
+                invalid_days = split == "days" and (
+                    leave.request_unit_hours or leave.request_unit_half
+                )
+
+                if invalid_hours or invalid_days:
+                    leave_sudo = leave.sudo().with_context(
+                        mail_activity_automation_skip=True
+                    )
```
GuillemCForgeFlow: why is the `sudo()` needed? it should come at least with an explanation (not self-explanatory)

hr_holidays_approvals/__manifest__.py
```diff
@@ -0,0 +1,24 @@
...
+    "name": "Time Off Approvals",
+    "summary": """
+        Allow configuring approval requirements separately for hour-based and
+        day-based time off requests.
+    """,
+    "category": "Human Resources/Time Off",
+    "version": "17.0.1.0.0",
+    "license": "AGPL-3",
+    "author": "ForgeFlow, Odoo Community Association (OCA)",
+    "mantainer": "ForgeFlow",
```
GuillemCForgeFlow:
```suggestion
    "maintainer": "ForgeFlow",
```
there's a typo here. also, the `maintainers` key is typically used to indicate a specific user who will be responsible for maintaining the module's code. see more info here: https://github.com/OCA/maintainer-tools/blob/master/template/module/__manifest__.py

hr_holidays_approvals/__manifest__.py
```diff
@@ -0,0 +1,24 @@
...
+    "website": "https://github.com/OCA/hr-holidays",
+    "contributors": [
+        "Guillermo Navas <guillermo.navas@forgeflow.com>",
+    ],
+    "depends": ["hr_holidays"],
+    "data": [
+        "views/hr_leave_type_views.xml",
+    ],
+    "installable": True,
+    "application": False,
```
GuillemCForgeFlow:
```suggestion
```
no need to add those every time, these are already the defaults

GuillemCForgeFlow (changes requested):

hr_holidays_approvals/models/hr_leave.py
```diff
@@ -0,0 +1,32 @@
+from odoo import api, models
+
+
+class HolidaysRequest(models.Model):
+    _inherit = "hr.leave"
+
+    @api.model_create_multi
+    def create(self, vals_list):
+        holidays = super(
+            HolidaysRequest, self.with_context(mail_create_nosubscribe=True)
```
GuillemCForgeFlow: why this context? I understand it has been copied from core, but there's no need to add it here again.

hr_holidays_approval_split/views/hr_leave_type_views.xml
GuillemCForgeFlow: can you answer this part? why was this done originally? was it on purpose?

hr_holidays_approval_split/models/hr_leave.py
```diff
@@ -0,0 +1,32 @@
...
+                    h.holiday_status_id.leave_validation_split == "hours"
+                    and not h.request_unit_hours
+                    and not h.request_unit_half
+                )
+                or (
+                    h.holiday_status_id.leave_validation_split == "days"
+                    and (h.request_unit_hours or h.request_unit_half)
+                )
+            )
+        )
```
GuillemCForgeFlow, line 26: you are not checking the `leave_validation_type` here. it should also be done for the cases where it is not `no_validation`, otherwise you'll repeat the call for `action_validate`

hr_holidays_approvals/models/hr_leave.py
```diff
@@ -0,0 +1,32 @@
...
+                    h.holiday_status_id.leave_validation_split == "days"
+                    and (h.request_unit_hours or h.request_unit_half)
+                )
+            )
+        )
+
+        if auto_validate:
+            # We use sudo() here because in the original hr.leave create method,
+            # action_validate is also called with sudo
+            auto_validate.sudo().action_validate()
```
GuillemCForgeFlow:
```suggestion
        if auto_validate:
            # As done in core, we use sudo() as the user might not have the rights when creating the leave
            auto_validate.sudo().action_validate()
```
instead of adding this comment for which you need to go to the odoo's code to check.
on top of that, as I see this is more or less what's done in core's:

```
if holiday.validation_type == 'no_validation':
    # Automatic validation should be done in sudo, because user might not have the rights to do it by himself
    holiday_sudo.action_validate()
    holiday_sudo.message_subscribe(partner_ids=holiday._get_responsible_for_approval().partner_id.ids)
    holiday_sudo.message_post(body=_("The time off has been automatically approved"), subtype_xmlid="mail.mt_comment") # Message from OdooBot (sudo)
```
why not adding the messages as well?

hr_holidays_approval_split/models/hr_leave.py
```diff
@@ -0,0 +1,32 @@
...
+class HolidaysRequest(models.Model):
+    _inherit = "hr.leave"
+
+    @api.model_create_multi
+    def create(self, vals_list):
+        holidays = super(
+            HolidaysRequest, self.with_context(mail_create_nosubscribe=True)
+        ).create(vals_list)
+
+        auto_validate = holidays.filtered(
```
GuillemCForgeFlow, line 11: to have this logic fully covered, I'd suggest adding some test files

hr_holidays_approvals/models/hr_leave_type.py
```diff
@@ -0,0 +1,19 @@
+from odoo import fields, models
+
+
+class HrLeaveType(models.Model):
+    _inherit = "hr.leave.type"
+
+    leave_validation_split = fields.Selection(
```
GuillemCForgeFlow: also, maybe the field name should follow the odoo's field convention, so we could rename to `validation_type_split`?

GuillemCForgeFlow (changes requested):

hr_holidays_approval_split/models/hr_leave.py
```diff
@@ -0,0 +1,40 @@
...
+            # As done in core, we use sudo() as the user might not have the
+            # rights when creating the leave
+            holiday.sudo().action_validate()
+            holiday.sudo().message_subscribe(
+                partner_ids=holiday._get_responsible_for_approval().partner_id.ids
+            )
+            holiday.sudo().message_post(
+                body=_("The time off has been automatically approved"),
+                subtype_xmlid="mail.mt_comment",
+            )
```
GuillemCForgeFlow, line 39: you can add the comment that this is based on odoo's core + add the link
this will help in the future, if the logic changes there

hr_holidays_approvals/__manifest__.py
```diff
@@ -0,0 +1,21 @@
+# Copyright 2026 ForgeFlow S.L. (https://www.forgeflow.com)
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
+{
+    "name": "Time Off Approvals",
```
GuillemCForgeFlow: instead of `hr_holidays_approvals` I would rename to `hr_holidays_approval_split` to be more specific

GuillemCForgeFlow (changes requested):
thanks for the tests 👍🏿

GuillemCForgeFlow (approved):
lgtm 👍🏿

PR #245, opened 2026-04-02
[18.0][ADD] hr_holidays_leave_report_calendar_type

dalonsod:
By default, general calendar view for leaves ("Overview") has no leave type filter.

This addon simply adds it to right filter panel, and at search bar as well.

imagen


dalonsod:
@rrebollo @bizzappdev could you review? Thanks!

rrebollo (approved):
Code Review. LGTM!

flotho (approved):
successfully tested on runboat.
LGTM
Thanks

IriaAlonso (commented):
LGTM only minor comment

hr_holidays_leave_report_calendar_type/__init__.py
```diff
@@ -0,0 +1 @@
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
```
IriaAlonso: This line is not needed here

hr_holidays_leave_report_calendar_type/views/hr_leave_type.xml
```diff
@@ -0,0 +1,12 @@
+<?xml version="1.0" encoding="UTF-8" ?>
```
DantePereyra: I think this is also unnecessary

hr_holidays_leave_report_calendar_type/views/hr_leave_type.xml
cgarcia-solvos, line 1: I think the views/ directory is unnecessary. Could you please verify? It should probably be removed

hr_holidays_leave_report_calendar_type/__init__.py
dalonsod: Comment addressed

hr_holidays_leave_report_calendar_type/views/hr_leave_type.xml
dalonsod: Comment addressed

hr_holidays_leave_report_calendar_type/views/hr_leave_type.xml
dalonsod: Realized that the file wasn't actually required, so it's bee removed

cgarcia-solvos (approved):

dalonsod:
@pedrobaeza is this ready to be merged? Thanks!

PR #248, opened 2026-04-09
[FIX] hr_holidays_natural_period: Create no_validation leave

SirPyTech:
**Steps**:
1. Create an employee that is not a HR manager
2. Create a leave type that counts natural days, but  no validation and no allocation
3. With the created employee's user, create a leave of the created leave type

**Actual**:
The following exception is raised:
<details><summary>Due to security restrictions, you are not allowed to modify 'Time Off' (hr.leave) records.</summary>
<pre>
ERROR devel odoo.addons.hr_holidays_natural_period.tests.test_hr_leave: ERROR: Subtest TestHrLeave.test_hr_leave_natural_day_no_manager (login='test-user')
Traceback (most recent call last):
  File "<decorator-gen-185>", line 2, in test_hr_leave_natural_day_no_manager
  File "/opt/odoo/custom/src/odoo/odoo/tests/common.py", line 2157, in _users
    func(*args, **kwargs)
  File "/opt/odoo/auto/addons/hr_holidays_natural_period/tests/test_hr_leave.py", line 174, in test_hr_leave_natural_day_no_manager
    leave = self._create_hr_leave(leave_type, "3000-01-01", "3000-01-01")
  File "/opt/odoo/auto/addons/hr_holidays_natural_period/tests/test_hr_leave.py", line 80, in _create_hr_leave
    return leave_form.save()
  File "/opt/odoo/custom/src/odoo/odoo/tests/common.py", line 2603, in save
    r = self._model.create(values)
  File "<decorator-gen-165>", line 2, in create
  File "/opt/odoo/custom/src/odoo/odoo/api.py", line 430, in _model_create_multi
    return create(self, [arg])
  File "/opt/odoo/auto/addons/hr_holidays_natural_period/models/hr_leave.py", line 29, in create
    )._compute_number_of_days()
  File "/opt/odoo/auto/addons/hr_holidays/models/hr_leave.py", line 512, in _compute_number_of_days
    holiday.number_of_days = holiday._get_number_of_days(holiday.date_from, holiday.date_to, holiday.employee_id.id)['days']
  File "/opt/odoo/custom/src/odoo/odoo/fields.py", line 1337, in __set__
    records.write({self.name: write_value})
  File "/opt/odoo/auto/addons/hr_holidays/models/hr_leave.py", line 1017, in write
    result = super(HolidaysRequest, self).write(values)
  File "/opt/odoo/auto/addons/mail/models/mail_thread.py", line 309, in write
    return super(MailThread, self).write(values)
[...]

PR #254, opened 2026-05-13
[18.0][FIX] hr_holidays_natural_period: Create no_validation leave

SirPyTech:
Forward port of https://github.com/OCA/hr-holidays/pull/248.
There is no change to the module's code because apparently this version is not affected by the same issue that was present in `16.0`.


quirino95 (approved):
Code and functional review: LGTM!

BhaveshHeliconia (approved):
LGTM!

hr_holidays_natural_period/tests/test_hr_leave.py
```diff
@@ -142,6 +151,32 @@ def test_hr_leave_natural_day_half_day_02(self):
...
+            self.env.ref("hr.group_hr_manager"),
+            self.env.ref("hr_holidays.group_hr_holidays_user"),
+            self.env.ref("hr_holidays.group_hr_holidays_responsible"),
+            self.env.ref("hr_holidays.group_hr_holidays_manager"),
+        ]
+        # pre-condition
+        self.assertEqual(leave_type.leave_validation_type, "no_validation")
+        for manager_group in manager_groups:
+            self.assertNotIn(manager_group, self.env.user.groups_id)
+
```
pedrobaeza, line 173: Please don't add empty lines inside a method.

SirPyTech: Hi, thanks for having a look!
Sometimes I use blank lines in a method to keep logical sections of the method separated, this kind of usage is even mentioned in PEP8 https://peps.python.org/pep-0008/#blank-lines:
> Use blank lines in functions, sparingly, to indicate logical sections.

Is there any specific reason why they should be removed? If that's just an opinion-based style decision I'd rather keep them.

pedrobaeza: Yes, because:

- `sparingly` means infrequent, not the usual thing to do.
- keep the current style of the module, which no lines inside the method.
- What is for you a logical section? It can be different from me, and adding a blank line doesn't enforce what that "logical section" does. If you really want to have logical sections, add a line with a comment explaining what the next logical section is.

hr_holidays_natural_period/tests/test_hr_leave.py
```diff
@@ -142,6 +151,32 @@ def test_hr_leave_natural_day_half_day_02(self):
...
+        # Act
+        leave = self._create_hr_leave(leave_type, "3000-01-01", "3000-01-01")
+
+        # Assert
```
pedrobaeza, line 177: This comment is redundant, as the Python code already infers that you assert in the next line.

hr_holidays_natural_period/tests/test_hr_leave.py
```diff
@@ -142,6 +151,32 @@ def test_hr_leave_natural_day_half_day_02(self):
...
+            self.env.ref("hr_holidays.group_hr_holidays_user"),
+            self.env.ref("hr_holidays.group_hr_holidays_responsible"),
+            self.env.ref("hr_holidays.group_hr_holidays_manager"),
+        ]
+        # pre-condition
+        self.assertEqual(leave_type.leave_validation_type, "no_validation")
+        for manager_group in manager_groups:
+            self.assertNotIn(manager_group, self.env.user.groups_id)
+
+        # Act
```
pedrobaeza, line 174: No need of this comment.

hr_holidays_natural_period/tests/test_hr_leave.py
SirPyTech: I see that https://github.com/OCA/hr-holidays/pull/254/changes#r3361035074 triggered more heated comments on the rest of the code :fire: sorry to have upset you.

Just to give you more context, this test method follows the AAA testing pattern (see for instance https://dev.to/coderjay06/the-three-a-s-of-unit-testing-b22) and most of the times I find it useful to give a clear structure to test methods:

> **A**rrange the state of the data to set it up for testing.
> 
> **A**ct on the data through some method that performs an action.
> 
> **A**ssert that the result from acting on that data is what we expect it to be.

The 3 logical sections in this method follow this structure.

hr_holidays_natural_period/tests/test_hr_leave.py
SirPyTech: This clarifies that we are in the second step of the AAA testing pattern (see for instance https://dev.to/coderjay06/the-three-a-s-of-unit-testing-b22).

hr_holidays_natural_period/tests/test_hr_leave.py
SirPyTech: This clarifies that we are in the third step of the AAA testing pattern (see for instance https://dev.to/coderjay06/the-three-a-s-of-unit-testing-b22).

SirPyTech (commented):
Apparently we have different opinions on using blank lines, but that happens, I hope this won't block the PR 🤞.

pedrobaeza:
My thumb rule is simple: less is more. To not have to do more vertical scrolling. All of this pattern thing is very good, but that's a methodology, not something that should be literally ported putting that comments. You don't even put `# AAA testing pattern: Assert` that at least will clarify more. You put `# Assert`. About that methodology, the arrange phase can't be always be done at the beginning, but let's not start that debate. 

The most important thing here is that this not a new module, just an added test, where the rest of the tests are not done that way, so please respect the current general style.

SirPyTech:
> My thumb rule is simple: less is more. To not have to do more vertical scrolling. All of this pattern thing is very good, but that's a methodology, not something that should be literally ported putting that comments. You don't even put `# AAA testing pattern: Assert` that at least will clarify more. You put `# Assert`. About that methodology, the arrange phase can't be always be done at the beginning, but let's not start that debate.
> 
> The most important thing here is that this not a new module, just an added test, where the rest of the tests are not done that way, so please respect the current general style.

The rest of the tests do not even have the docstring, but the new test has it and that does not look to be a problem meaning improvement can be done even if means a difference from the other tests, and I think separating the logical blocks is an improvement.

Anyway, I have removed the blank lines to reduce the vertical scrolling.
I like it a bit less, and you probably like it a bit more.
Can this be considered a common ground so this can be merged?
Thanks!

SirPyTech:
@victoralmau as the module maintainer, what do you think of this one?

hr_holidays_natural_period/tests/test_hr_leave.py
```diff
@@ -142,6 +151,30 @@ def test_hr_leave_natural_day_half_day_02(self):
...
+            self.env.ref("hr_holidays.group_hr_holidays_manager"),
+        ]
+        # pre-condition
+        self.assertEqual(leave_type.leave_validation_type, "no_validation")
+        for manager_group in manager_groups:
+            self.assertNotIn(manager_group, self.env.user.groups_id)
+        # Act
+        leave = self._create_hr_leave(leave_type, "3000-01-01", "3000-01-01")
+        # Assert
+        self.assertEqual(leave.state, "validate")
```
victoralmau:
```suggestion
        self.leave_type.requires_allocation = "no"
        self.leave_type.leave_validation_type = "no_validation"        
        leave = self._create_hr_leave(leave_type, "3000-01-01", "3000-01-01")
        self.assertEqual(leave.state, "validate")
```

hr_holidays_natural_period/tests/test_hr_leave.py
```diff
@@ -15,6 +17,13 @@ def setUpClass(cls):
         cls.leave_type = cls.env.ref(
             "hr_holidays_natural_period.hr_leave_type_natural_day_test"
         )
+        cls.no_validation_leave_type = cls.leave_type.copy(
+            default={
+                "name": "Test no validation leave type",
+                "requires_allocation": "no",
+                "leave_validation_type": "no_validation",
+            },
+        )
```
victoralmau:
```suggestion
```

victoralmau (changes requested):
These suggestions can also be added to https://github.com/OCA/hr-holidays/pull/255

If it turns out that no additional changes are needed in `hr.leave`, I think the commit should be IMP and should mention adding a use case with no_validation.

hr_holidays_natural_period/tests/test_hr_leave.py
SirPyTech: ok we don't need the additional leave type, but we have to ensure that the current user is not a manager, don't you think?
Both the test signature and docstring highlight that is the use-case being tested.

hr_holidays_natural_period/tests/test_hr_leave.py
SirPyTech: :+1: removed

hr_holidays_natural_period/tests/test_hr_leave.py
victoralmau: None of that is necessary because we're using `@users(“test-user”)`, a user we created on line 42 (setup) who doesn't have any of those permissions.

SirPyTech: Right, I did prefer to check for that pre-condition explicitly but I understand it can look redundant; we'll have to hope no-one will add manager groups to the test user :man_shrugging:

I removed the comments too because they were almost doubling the method length at this point :sweat_smile:

I also renamed the commit, please check again

victoralmau (approved):

pedrobaeza (changes requested):
I would argue about adding as contributor for such tiny test method.

SirPyTech:
> I would argue about adding as contributor for such tiny test method.

Contribution is not only measured in lines of code; by the way I wrote many lines.. of issue comments :smile:

I also tested the module both here and in the other versions this change involved and made the needed adaptations so why shouldn't I be among the contributors?
In the end a few lines are needed, but for sure I spent a lot of time for its improvement.

pedrobaeza (approved):
OK, I don't want to fight more, but it's really displeasant and exhausting to review any of your contributions.

/ocabot merge patch

SirPyTech:
> OK, I don't want to fight more, but it's really displeasant and exhausting to review any of your contributions.
> 
> /ocabot merge patch

It happens to have different opinions :man_shrugging: when it happens we have to accept that some sort of compromise has to be found on both sides, I'm sorry it is displeasing.

Thanks for merging!

pedrobaeza:
The problem is not to have different opinions, but to having to expose reasonings for every change requested while most of them are styles and these are modules with a determined style.

SirPyTech:
I don't think that happens in *any of my contributions*: it happened here
and a few other times.
And that happens exactly because we have different opinions *on style*.
Exposing the reasoning when asking for a change should be the usual
behavior when reviewing other people's work, I'm sorry it is displeasing
for you, but is part of the reviewer's job.

The problem is when Pull Requests get blocked for these kind of problems
because they are not problems.
As you can see I have strong opinions on style, like you, but when I review
other Pull Requests I don't block them because of style issues: sure I
write them down but I make it clear that they are not blocking.

Il giorno lun 14 set 2026 alle ore 16:36 Pedro M. Baeza <
***@***.***> ha scritto:

pedrobaeza:
The problem is to have a module with mixed styles, and that's what I try to avoid. Anyway, let's reset and go for the next contribution, as I think you are a good contributor no matter all of these little discussions.

PR #267, opened 2026-07-23
[17.0][IMP] `hr_holidays_public`: improve way to get public holidays

Camille0907:
Superseeds https://github.com/OCA/hr-holidays/pull/173

This first commit is exactly the same than the original PR.

---

[FIX] hr_holidays_public: fix AttributeError in _attendance_intervals_batch

Previous commit added `resources.mapped("employee_id")` to auto-set `employee_id` in context.
However, [PR OCA#225](OCA#225) had previously set `resources` to a
Python list `[self.env["resource.resource"]]` for the `resources=None` case, which
has no `.mapped()` method, causing an AttributeError whenever
`exclude_public_holidays` was set in context without explicit resources.

Fix by adopting the same `resources_list` pattern used in Odoo core
`_attendance_intervals_batch`: when `resources` is None, use a single-element list
containing the empty recordset (key `False`), otherwise extend the resource list
with the empty recordset. Pass `resources_list` to the exclude helper so that
`intervals[False]` is always filtered, preserving the feature from PR OCA#225.


hr_holidays_public/models/resource_calendar.py
```diff
@@ -46,9 +46,17 @@ def _attendance_intervals_batch(
             lunch=lunch,
         )
         if resources is None:
-            resources = [self.env["resource.resource"]]
-        if self.env.context.get("exclude_public_holidays") and resources:
+            resources = self.env["resource.resource"]
+            resources_list = [resources]
+        else:
+            resources_list = list(resources) + [self.env["resource.resource"]]
+        if self.env.context.get("exclude_public_holidays"):
```
SilvioC2C, line 53: The previous condition was `if self.env.context.get("exclude_public_holidays") and resources:`, but now the code only checks for the context key. Is this change needed/wanted?

hr_holidays_public/models/resource_calendar.py
```diff
@@ -46,9 +46,17 @@ def _attendance_intervals_batch(
...
+            # We need to pass the employee_id to the context to be able to filter
+            # the public holidays correctly
+            employee_ids = resources.mapped("employee_id")
+            if len(employee_ids) == 1:
+                self = self.with_context(employee_id=employee_ids.id)
```
SilvioC2C: Nitpicking:
- `mapped` is not needed to read a relational field on a recordset (at least since Odoo 14.0)
- `employees` would be a better name than `employee_ids`, since `resources.employee_id` would return a recordset, not their IDs

SilvioC2C (approved):
Pre-approving

ivantodorovich (approved):

vvrossem (approved):

hr_holidays_public/models/resource_calendar.py
Camille0907: Yes, in fact when you look at the previous code, if resources is None, we replace it by a truthy value, so `and resources` was always `and True`.
In my code `resources` can be falsy, so to keep the same behavior as before I removed the `and resources`

hr_holidays_public/models/resource_calendar.py
Camille0907: True, I cherry-picked the commit without using my brain 😅 I'll change

BhaveshHeliconia (approved):

PR #269, opened 2026-08-07
[19.0][ADD] hr_holidays_leave_type_visibility

JasminSForgeFlow:
## Purpose

In standard Odoo, the *Time Off Type* of an absence shown in the Time Off
Overview calendar (`hr.leave.report.calendar`) is restricted to the
*Time Off / Officer: Manage all requests* group. Regular employees can therefore
see **that** a colleague is absent, but not **why**.

Several customers need employees to know the kind of absence (sick leave,
holiday, training, ...) in order to plan their own work, without granting anyone
approval or write rights on time off requests.

@ForgeFlow


AaronHForgeFlow (commented):
In runboat I got this error when trying to create a leave for the demo user:

Image

Is that due to this module or another? can you check?

JasminSForgeFlow:
Tests fixed here https://github.com/OCA/hr-holidays/pull/270

JasminSForgeFlow:
> In runboat I got this error when trying to create a leave for the demo user:
> 
> Image
> Is that due to this module or another? can you check?

This error is not caused by this module — it reproduces on the same runboat with `hr_holidays_leave_type_visibility` uninstalled.

**Root cause (upstream `hr_holidays`):** in Odoo 19 `hr.employee` delegates to `hr.version` via `_inherits`, and the link field `version_id` is declared with `groups="hr.group_hr_user"` (`addons/hr/models/hr_employee.py`). So any read of a delegated field on `hr.employee` by a plain internal user raises the "Implicitly accessed through 'Employee'" AccessError.

In the leave-creation flow that happens in two places, both reading `employee_id.department_id` (defined on `hr.version`, `addons/hr/models/hr_version.py`) without `sudo()`:

- `HrLeave._compute_department_id` — `addons/hr_holidays/models/hr_leave.py:537`
- `HrLeave._compute_has_mandatory_day` — `addons/hr_holidays/models/hr_leave.py:549`

Note that the neighbouring lines in the same file *do* sudo for exactly this reason, with explanatory comments:

- `hr_leave.py:543` — `# Sudo to get access to version fields on employee (job_id)`
- `hr_leave.py:623` — `# sudo as is_flexible is on version model and employee does not have access to it.`

So it looks like an inconsistency in core rather than intended behaviour.

**Verified on runboat:**

- demo user without `hr.group_hr_user` → AccessError on `version_id`, with or without this module installed
- demo user granted `hr.group_hr_user` → leave creates fine

This module only touches `hr.leave.report.calendar` (widening `holiday_status_id` to `base.group_user` and overriding `_compute_name`); it never reads employee version fields. The fix belongs in odoo/odoo, not here.

hr_holidays_leave_type_visibility/tests/test_leave_type_visibility.py
```diff
@@ -0,0 +1,79 @@
+# Copyright 2026 ForgeFlow S.L.
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
+
+from datetime import date
+
+from odoo.tests.common import TransactionCase, new_test_user
+
+
+class TestLeaveTypeVisibility(TransactionCase):
```
BhaveshHeliconia, line 9:
```suggestion
class TestLeaveTypeVisibility(Basecommon):
```

BhaveshHeliconia (approved):
Code review LGTM!

Just minor suggestion

PR #201, opened 2025-07-15
[16.0][ADD] hr_holidays_accrual_by_tag: new module

quirino95:
This module automates creation and management of accrual allocations for employees, based on assigned tags (which have to be related to an accrual plan).

When a new employee is created, and some tags are assigned to, related accrual allocations are created automatically when saving.
Also adding a tag to an existing employee will create an accrual allocation.
If a tag is removed from an employee, `date_to` on related accrual allocation is written in order to stop accruing hours.
If the same tag is added again, a new independent accrual allocation is created.

Of course, is already possible to do all these steps manually, but it' very hard. Using this module makes everything easier.

From a technical point of view, in order to achieve this, I've extended behavior of `create` and `write` methods of `hr.employee` module.

PR #276, opened 2026-09-08
[ADD] hr_holidays_accrual_by_tag

SirPyTech:
Proposing again the module that is still being proposed for `16.0` in https://github.com/OCA/hr-holidays/pull/201.

The migration commit partially rewrites the logic according to the new behavior defined in https://github.com/odoo/odoo/pull/158829.


hr_holidays_accrual_by_tag/models/hr_employee.py
```diff
@@ -0,0 +1,68 @@
...
+                ("generate_allocation_category_ids", "!=", False),
+            ]
+        )
+        existing_assignments = self.env["hr.leave.allocation"].search(
+            [
+                ("employee_id", "in", self.ids),
+                ("accrual_plan_id", "in", accrual_plans.ids),
+                ("date_to", "=", False),
+            ],
+        )
```
HekkiMelody, line 35: **issue**: to better filter existing allocations, we need to filter for `("state", "=", "confirm")` as well.
Date to could be `False` or `("date_to", ">", fields.Date.context_today())` to account for the user making changes to the allocations manually

hr_holidays_accrual_by_tag/models/hr_leave_accrual_plan.py
```diff
@@ -0,0 +1,23 @@
...
+
+class HrLeaveAccrualPlan(models.Model):
+    _inherit = "hr.leave.accrual.plan"
+
+    generate_allocation_category_ids = fields.Many2many(
+        comodel_name="hr.employee.category",
+        relation="hr_leave_accrual_category_rel",
+        string="Generate allocations by tag",
+        help="When one of these tags is added to an employee: "
+        "an accrual allocation for the employee created.\n"
```
HekkiMelody: **typo**: 

```suggestion
        "an accrual allocation for the employee is created.\n"
```

HekkiMelody (changes requested):
Functional review, works well. Let's clean up the edge cases and we should be good!

hr_holidays_accrual_by_tag/models/hr_leave_accrual_plan.py
SirPyTech: 👍 thanks for catching this

hr_holidays_accrual_by_tag/models/hr_employee.py
SirPyTech: It makes sense 👍

New allocations are automatically validated so I had to filter for `validate` status instead, please let me know if this is ok, otherwise I'd like to know more about the use-case you're trying to fix.
I added a test case for what I think the use-case was.

HekkiMelody: `validate` was the correct status, thank you!

HekkiMelody (approved):
Code and functional review, LGTM

hr_holidays_accrual_by_tag/models/hr_employee.py
```diff
@@ -0,0 +1,70 @@
...
+        return res
+
+    def _assign_accrual_plans_by_tags(self):
+        today = fields.Date.context_today(self)
+        accrual_plans = self.env["hr.leave.accrual.plan"].search(
+            [
+                ("generate_allocation_category_ids", "!=", False),
+            ]
+        )
+        existing_assignments = self.env["hr.leave.allocation"].search(
```
BhaveshHeliconia:
```suggestion
        if not accrual_plans:
            return
        existing_assignments = self.env["hr.leave.allocation"].sudo().search(
```

hr_holidays_accrual_by_tag/models/hr_employee.py
```diff
@@ -0,0 +1,70 @@
...
+                                "accrual_plan_id": plan.id,
+                            }
+                        )
+                    elif category not in employee_tags and employee_assignment:
+                        employee_assignment.date_to = today
+
+        if generate_new_allocations_values:
+            generate_new_allocations = self.env[
+                "hr.leave.allocation.generate.multi.wizard"
+            ].create(generate_new_allocations_values)
```
BhaveshHeliconia:
```suggestion
            ]sudo().create(generate_new_allocations_values)
```
A standard HR Officer (hr.group_hr_user) can create and update employees, but the allocation generation wizard (hr.leave.allocation.generate.multi.wizard) is restricted to Time Off Officers (hr_holidays.group_hr_holidays_user).

Without .sudo(), adding a tag to an employee as a regular HR Officer triggers an AccessError because the wizard cannot be accessed. Using .sudo() lets the automated hook run regardless of which user is editing the employee.

Also, setting "company_id": employee.company_id.id ensures the wizard uses the employee’s company calendar correctly in a multi-company setup.

hr_holidays_accrual_by_tag/models/hr_employee.py
```diff
@@ -0,0 +1,70 @@
...
+                            {
+                                "allocation_mode": "employee",
+                                "allocation_type": "accrual",
+                                "holiday_status_id": holiday_status.id,
+                                "employee_ids": employee.ids,
+                                "accrual_plan_id": plan.id,
+                            }
+                        )
+                    elif category not in employee_tags and employee_assignment:
+                        employee_assignment.date_to = today
```
BhaveshHeliconia:
```suggestion
                has_matching_tag = bool(
                    employee_tags & plan.generate_allocation_category_ids
                )
                if has_matching_tag and not employee_assignment:
                    generate_new_allocations_values.append(
                        {
                            "allocation_mode": "employee",
                            "allocation_type": "accrual",
                            "holiday_status_id": holiday_status.id,
                            "employee_ids": employee.ids,
                            "accrual_plan_id": plan.id,
                            "company_id": employee.company_id.id,
                        }
                    )
                elif not has_matching_tag and employee_assignment:
                    employee_assignment.date_to = max(
                        today, employee_assignment.date_from
                    )
```
Duplicate allocations: When a plan is linked to multiple tags, and an employee has more than one of those tags, the old code could create the same allocation more than once. For example, with [Full-Time, Senior], both tags could trigger a new allocation.

Premature cancellation: The same logic could also end an existing allocation too early. If the employee still had one matching tag but not the other, the allocation could be marked as ended even though the employee was still eligible.

Fix: The new logic checks all matching tags together instead of handling them one by one. The allocation is created only once if the employee has at least one matching tag, and it is only ended when none of the plan's tags are left on the employee.

Odoo does not allow date_from to be later than date_to. If an employee has a future start date, setting date_to to today can cause a ValidationError. Using max(today, employee_assignment.date_from) prevents this invalid date range.

hr_holidays_accrual_by_tag/views/hr_leave_accrual_plan_views.xml
```diff
@@ -0,0 +1,29 @@
...
+                    <field
+                        name="generate_allocation_category_ids"
+                        string="Tags"
+                        widget="many2many_tags"
+                    />
+                    <field
+                        name="generate_allocation_default_status_id"
+                        string="Default Time Off Type"
+                        invisible="not generate_allocation_category_ids"
+                        required="generate_allocation_category_ids"
```
BhaveshHeliconia, line 23:
```suggestion
                        required="generate_allocation_category_ids"
                        domain="['|', ('id', '=', time_off_type_id), ('time_off_type_id', '=', False)]"
```

hr_holidays_accrual_by_tag/__manifest__.py
```diff
@@ -0,0 +1,23 @@
...
+    "author": "PyTech SRL, Odoo Community Association (OCA)",
+    "maintainers": [
+        "HekkiMelody",
+        "SirPyTech",
+    ],
+    "website": "https://github.com/OCA/hr-holidays",
+    "depends": [
+        "hr",
+        "hr_holidays",
+    ],
```
BhaveshHeliconia, line 18:
```suggestion
    "depends": [
        "hr_holidays",
    ],
```

BhaveshHeliconia:
Suggestion: We should keep generate_allocation_default_status_id in sync with time_off_type_id. Otherwise, users can select different leave types and the allocation will fail later with a ValidationError.

An onchange, constraint, and view domain could prevent this and make the behavior clearer.

hr_holidays_accrual_by_tag/models/hr_employee.py
SirPyTech: Early exit + `sudo` for the HR/Time Off Officer issue :+1:

hr_holidays_accrual_by_tag/models/hr_employee.py
SirPyTech: Makes sense, added :+1:

hr_holidays_accrual_by_tag/models/hr_employee.py
SirPyTech: And we remove a nested `for` loop, thank you for your suggestion :pray:

hr_holidays_accrual_by_tag/views/hr_leave_accrual_plan_views.xml
SirPyTech: This required some fixing, I used `=?`, please check if that's ok.

hr_holidays_accrual_by_tag/__manifest__.py
SirPyTech: Streamlined the dependencies :+1:

SirPyTech (commented):
Thanks @BhaveshHeliconia for your review!
You raised many valid points, please check if that's better now, thanks!

hr_holidays_accrual_by_tag/models/hr_employee.py
```diff
@@ -0,0 +1,80 @@
...
+                    ("date_to", "=", False),
+                    ("date_to", ">", today),
+                ],
+            )
+        )
+
+        generate_new_allocations_values = []
+        for employee in self:
+            employee_tags = employee.category_ids
+            for plan in accrual_plans:
```
BhaveshHeliconia, line 51:
```suggestion
            for plan in accrual_plans:
                if plan.company_id and plan.company_id != employee.company_id:
                    continue
```

hr_holidays_accrual_by_tag/models/hr_employee.py
```diff
@@ -0,0 +1,80 @@
...
+                    employee_tags & plan.generate_allocation_category_ids
+                )
+                if has_matching_tag and not employee_assignment:
+                    generate_new_allocations_values.append(
+                        {
+                            "allocation_mode": "employee",
+                            "allocation_type": "accrual",
+                            "holiday_status_id": holiday_status.id,
+                            "employee_ids": employee.ids,
+                            "accrual_plan_id": plan.id,
```
BhaveshHeliconia, line 74:
```suggestion
                            "accrual_plan_id": plan.id,
                            "company_id": employee.company_id.id,
```
if plan.company_id and plan.company_id != employee.company_id: continue prevents trying to assign a plan from Company A to an employee in Company B.

Adding "company_id": employee.company_id.id ensures the wizard generates allocations under the employee's company instead of defaulting to the logged-in user's active company (self.env.company).

hr_holidays_accrual_by_tag/models/hr_leave_accrual_plan.py
```diff
@@ -0,0 +1,45 @@
...
+        string="Generate allocations by tag",
+        help="When one of these tags is added to an employee: "
+        "an accrual allocation for the employee is created.\n"
+        "When one of these tags is removed from an employee: "
+        "the corresponding accrual allocation is ended.\n",
+    )
+    generate_allocation_default_status_id = fields.Many2one(
+        comodel_name="hr.leave.type",
+        string="Default Time Off Type for generated allocations",
+        help="Type of the generated allocations.",
```
BhaveshHeliconia, line 23:
```suggestion
        help="Type of the generated allocations.",
        check_company=True,
```

hr_holidays_accrual_by_tag/views/hr_leave_accrual_plan_views.xml
```diff
@@ -0,0 +1,32 @@
...
+    ~ Copyright 2026 Simone Rubino - PyTech
+    ~ License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
+-->
+<odoo>
+    <record id="hr_accrual_plan_view_form" model="ir.ui.view">
+        <field name="name">Add Accrual by Tag fields to Accrual Plan form view</field>
+        <field name="model">hr.leave.accrual.plan</field>
+        <field name="inherit_id" ref="hr_holidays.hr_accrual_plan_view_form" />
+        <field name="arch" type="xml">
+            <xpath expr="//sheet/group" position="inside">
```
BhaveshHeliconia, line 12:
```suggestion
            <xpath expr="//sheet/group" position="inside">
                <field name="time_off_type_id" invisible="1" />
```
ensures the web client loads the field so the dynamic domain evaluates properly in the ui.

hr_holidays_accrual_by_tag/tests/test_hr_holidays_accrual_by_tag.py
```diff
@@ -0,0 +1,136 @@
+from odoo import Command
+from odoo.tests.common import TransactionCase
+
+
+class TestHrHolidaysAccrualsByTag(TransactionCase):
```
BhaveshHeliconia:
```suggestion
class TestHrHolidaysAccrualsByTag(Basecommon):
```

BhaveshHeliconia (commented):
just few suggestions

hr_holidays_accrual_by_tag/models/hr_employee.py
SirPyTech: 👍

hr_holidays_accrual_by_tag/models/hr_employee.py
SirPyTech: :+1: it makes sense now that we have `sudo` because of the HR/Time Off Officer improvement

hr_holidays_accrual_by_tag/models/hr_leave_accrual_plan.py
SirPyTech: :+1: aligned with `time_off_type_id` of this model.

hr_holidays_accrual_by_tag/views/hr_leave_accrual_plan_views.xml
SirPyTech: This does not seem to be needed, could you please double check?

hr_holidays_accrual_by_tag/tests/test_hr_holidays_accrual_by_tag.py
SirPyTech: :+1: included in the migration

SirPyTech (commented):
Thank you @BhaveshHeliconia for keeping an eye on permission/multi-company aspects, in our simple use case we didn't need them so they were probably overlooked.

hr_holidays_accrual_by_tag/views/hr_leave_accrual_plan_views.xml
BhaveshHeliconia: the invisible field in the view is not necessary

BhaveshHeliconia (approved):
LGTM!

Thank you for addressing all the feedback and improving

HekkiMelody (approved):

PR #279, opened 2026-09-16
[18.0][IMP] hr_holidays_public: Use the employee's company partner if the employee does not have an address

victoralmau:
Use the employee's company partner if the employee does not have an address.

We will use the address of the employee or the partner of the company; the goal is to check only the public holidays for a specific country.

@Tecnativa


pedrobaeza (commented):
This has a side effect over existing base, as something that works previously (an employee without address, but having public holidays only for a country), now it stops working. What do you think about using the country of the employee's company as fallback instead?

pedrobaeza (approved):

carlos-lopez-tecnativa (approved):

PR #280, opened 2026-09-16
[19.0][IMP] hr_holidays_public: Use the employee's company partner if the employee does not have an address

victoralmau:
FWP from 18.0: https://github.com/OCA/hr-holidays/pull/279

Use the employee's company partner if the employee does not have an address.

We will use the address of the employee or the partner of the company; the goal is to check only the public holidays for a specific country.

@Tecnativa


pedrobaeza (approved):
/ocabot merge patch

